PR #1, opened 2022-06-23
feat: add rfc for bearer tokens

stringhandler:
An RFC for DAN layer bearer tokens

### DIFF
--- a/src/RFC_0388_BearerTokens.md
+++ b/src/RFC_0388_BearerTokens.md
@@ -1,0 +1,206 @@
+# RFC-0388 Bearer Tokens
+
+## A Scheme for Granting the Bearer Permissions on Second Layer Assets
+
+![status: raw](theme/images/status-raw.svg)
+
+**Maintainer(s)**: @mikethetike
+
+# Licence
+
+[ The 3-Clause BSD Licence](https://opensource.org/licenses/BSD-3-Clause).
+
+Copyright 2022 The Tari Development Community
+
+Redistribution and use in source and binary forms, with or without modification, are permitted provided that the
+following conditions are met:
+
+1. Redistributions of this document must retain the above copyright notice, this list of conditions and the following
+   disclaimer.
+2. Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following
+   disclaimer in the documentation and/or other materials provided with the distribution.
+3. Neither the name of the copyright holder nor the names of its contributors may be used to endorse or promote products
+   derived from this software without specific prior written permission.
+
+THIS DOCUMENT IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS", AND ANY EXPRESS OR IMPLIED WARRANTIES,
+INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+SERVICES; LOSS OF USE, DATA OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+WHETHER IN CONTRACT, STRICT LIABILITY OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF
+THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+## Language
+
+The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", 
+"NOT RECOMMENDED", "MAY" and "OPTIONAL" in this document are to be interpreted as described in 
+[BCP 14](https://tools.ietf.org/html/bcp14) (covering RFC2119 and RFC8174) when, and only when, they appear in all capitals, as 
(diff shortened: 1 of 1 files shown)


src/RFC_0388_BearerTokens.md
```diff
@@ -0,0 +1,206 @@
...
+| Field | Type | Description |
+| --- | --- | --- |
+| root_nonce | bytes(32) (optional) | a base ID used to revoke permissions(see below)  | 
+| expires_at_height | u64 (optional) | the sidechain height at which this token expires. Timestamps are not reliable in blockchains, so height is used in this case |
+| granted_to | pubkey | the public key of the grantee. Any bearer using this token will need to prove knowledge of the private key |
+| scopes | string[] | A list of scopes granted to this scope |
+| caveats | CaveatExpression[] | An ordered list of caveats |
+| based_on | Token (optional) | If this token is derived from another token, it should be present here |
+| issuer | PubKey | the issuer of this token |
+| issuer_sig | Signature | a signature of the challenge `Hash(base_id + granted_to + scopes + caveats + expires_at_height + based_on + issuer )` |
```
CjS77, line 82:
```suggestion
| issuer_sig | Signature | a signature signed by `issuer` of the challenge `Hash(root_nonce + granted_to + scopes + caveats + expires_at_height + based_on + issuer )` |
```

src/RFC_0388_BearerTokens.md
```diff
@@ -0,0 +1,206 @@
...
+
+### Caveat Expression
+
+``` 
+Caveat Expression = <Field> <operator> <Argument>
+```
+
+Field: An arbitrary string that will be interpreted by the code
+Operator: OneOf("eq", "le", "lt","ge", "gt")
+Argument: A constant value, to be interpretted by the code
```
CjS77, line 92:
```suggestion
Argument: A constant value, to be interpreted by the code
```

src/RFC_0388_BearerTokens.md
```diff
@@ -0,0 +1,206 @@
...
+
+The root_nonce, if specified must match the root nonce on record for the issuer.
+> Open question: should root nonce just be a special caveat?
+
+Before starting execution of the instruction, the list of auth tokens must be validated to ensure that each token is more restrictive than
+the last.
+
+When executing instructions, caveats MUST be checked if they are relevant to the resources being acted upon. 
+For example, if a function requires a scope `transfer`, the most specific AuthToken must have that scope present.
+
```
CjS77, line 115: Where and how does one gain access to these resources? This isn't clear.
How do we verify that the resources are the correct ones?

src/RFC_0388_BearerTokens.md
```diff
@@ -0,0 +1,206 @@
...
+
+
+## Revoking Tokens
+Each contract SHOULD store a root nonce for each identity in the contract. To revoke a set of tokens, an identity owner
+may change their root nonce. This will revoke all tokens based on this root.
+
```
CjS77, line 121: To clearly distinguish between asset tokens in the contract and authority bearer tokens, let's use a different word than "token" for the auth tokens.

Options could be:
- meringues (similar to macaroons)
- badges (implies authority)
- permits
- mandate
- accreditation
-

src/RFC_0388_BearerTokens.md
```diff
@@ -0,0 +1,206 @@
...
+```
+fn transfer(amount: u64, from: PublicKey, to: PublicKey) 
+{
+   requires_scope("transfer", from);
+   ...
+}
+```
+
+Alice creates a token:
+```json
```
CjS77, line 138:
```suggestion
```json5
```

src/RFC_0388_BearerTokens.md
```diff
@@ -0,0 +1,206 @@
...
+   "issuer_sig": "<sig>"
+}
+```
+
+> Note: The `root_nonce` and `expires_at_height` are missing here, but it would have been better for Alice to include these.
+> Also, this token allows Bob to spend up to 100 at a time, but does not restrict Bob from using this token again
+
+Bob can now create the instruction and submit it to the validator node. The validator node checks that the signature
+matches the public key in `granted_to` and also checks that the `amount` parameter is less than or equal to 100. Finally, the
+validator node checks that the scope includes 'transfer' and that Alice's public key is set in `from`.
```
CjS77, line 156:
```suggestion
Bob can now create the instruction and submit it to the validator node. The validator node checks that the instruction signature
matches the public key in `granted_to` and also checks that the `amount` parameter is less than or equal to 100. Finally, the
validator node checks that the scope includes 'transfer', and Alice's public key (specified in `from`) matches the auth token's issuer signature.
```

src/RFC_0388_BearerTokens.md
```diff
@@ -0,0 +1,206 @@
...
+
+### Example 2: Delegation
+Let's continue the example, but in this case Bob wants to allow Carol to spend the funds.
+
+In this case, Bob creates a token for Carol with the following:
+
+```json
```
CjS77, line 163:
```suggestion
```json5
```

CjS77 (commented):
Great RFC. Some minor editorial comments. Otherwise LGTM

sdbondi (dismissed):
Looks good! 

One important question that was alluded to but not answered was how to limit the times a particular bearer token can be used. For e.g. with the caveat `amount le 100` would you be permitted to continuously take 100 using that token? I suppose the granter may first allocate the funds to a separate pool/bucket and grant access to that, but this is similar to stateful on-chain approval. 

Perhaps a bearer token can optionally be linked to some on-chain token or account, and usage (and other) state can be tracked that way, similar to how some other cryptos have made authorisation tokens = NFTs.

src/RFC_0388_BearerTokens.md
```diff
@@ -0,0 +1,206 @@
...
+For example, the user may approve a service to transfer all of their funds, or all of their NFTs.
+
+The scheme proposed by this RFC is a stateless token that the bearer can use to invoke methods on the 
+DAN layer.
+
+## Related Requests for Comment
+
+## Description
+
+This scheme is inspired by [Macaroons] in that it allows the bearer of a token to delegate a more restrictive token to
```
CjS77, line 65:
```suggestion
This scheme is inspired by [Macaroons](https://www.ndss-symposium.org/ndss2014/programme/macaroons-cookies-contextual-caveats-decentralized-authorization-cloud/) in that it allows the bearer of a token to delegate a more restrictive token to
```

PR #2, opened 2022-06-28
docs: template registration UTXO

sdbondi:
Description
----
Adds the template registration UTXO RFC

Motivation and Context
----

Details how DAN templates could be registered on the base layer.

### DIFF
--- a/src/RFC-0306_DANTemplateRegistration.md
+++ b/src/RFC-0306_DANTemplateRegistration.md
@@ -83,31 +83,30 @@ are running the same code as other members of the same committee.
 
 A base-layer template registration [UTXO] MUST have the `TEMPLATE_REGISTRATION` output flag set and contain the following data:
 
-| Name              | Type               | Description                                                                                                                           |
-|-------------------|--------------------|---------------------------------------------------------------------------------------------------------------------------------------|
-| author_public_key | ECC public key     | Public key of the author                                                                                                              |
-| author_signature  | Schnorr signature  | Signature that signs remaining fields                                                                                                 |
-| template_name     | String (255)       | A descriptive name for the template                                                                                                   |
-| template_version  | Varint             | Code version as a single number                                                                                                       |
-| build_info        | BuildInfo struct   | Information on the binary build                                                                                                       |                                                                                                                             |
-| binary_checksum   | SHA2 checksum      | A SHA2 checksum of the WASM binary.                                                                                                   |
-| binary_address    | [Multiaddr]        | A [multiaddr] pointing to the WASM module binary download. This may be an HTTP, ONION, p2p, or IPFS address. Maximum byte length: 255 |
+| Name              | Type              | Description                                                                                                                             |
+|-------------------|-------------------|-----------------------------------------------------------------------------------------------------------------------------------------|
+| author_public_key | ECC public key    | Public key of the author                                                                                                                |
+| author_signature  | Schnorr signature | Signature that signs remaining fields                                                                                                   |
+| template_id       | 256-bit hash      | Hash of `TEMPLATE_REGISTRATION` fields (see below)                                                                                      |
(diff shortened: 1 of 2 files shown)


src/RFC-0306_DANTemplateRegistration.md
```diff
@@ -0,0 +1,173 @@
...
+
+The `author_signature` is a Schnorr proof that commits to the template fields contained in the [UTXO], namely
+`template_name`, `template_version`, `build_info`, `binary_checksum`, and `binary_address`.
+
+The base node acts as a notary for this data, it is not responsible for the validity of the template fields. However, it must
+not allow malformed/invalid data to be committed to the blockchain.
+
+Therefore, some additional base-layer consensus rules are required for a `TEMPLATE_REGISTRATION` [UTXO]:
+* A base node MUST validate the `author_signature`; and
+* The [UTXO] MUST have a relative time-lock of 100 blocks.
```
SWvheerden, line 104: What's the reason for the lock height?
I can think of reasons why you might want to push an updated version shortly. For example, you realized there is a bug in the just pushed version, now you have to wait for 200 minutes before you can push your fix?

src/RFC-0306_DANTemplateRegistration.md
```diff
@@ -0,0 +1,173 @@
...
+* Any live contracts SHOULD upgrade their definitions to run the new template.
+2. The author MAY spend to a "vanilla" [UTXO] to reclaim their deposit.
+* This effectively withdraws (yanks) the template.
+* The template MAY remain in use on existing contracts. In fact, anyone may now re-register the template.
+
+Alternatives:
+* Consensus rule that prevents spending of the template while used by other contracts
+* The contract definition may have copy the template binary_checksum etc. Validator nodes may mirror the code for new VNs
+
+## Upgrading a Template
```
SWvheerden, line 161: We should spec this out more as this could be troublesome if they upgrade out of sync

SWvheerden (commented):
This looks good, only have one question, and spec for a future PR

src/RFC-0306_DANTemplateRegistration.md
sdbondi: Yeah this was chosen pretty arbitrarily. I think it's more around having some assurance for this UTXO existing on the blockchain when referencing this template in a contract - which this doesn't really answer so I'll remove this.

Will have to put more thought into how we reference template UTXOs or if we copy the template data into the contract definition/constitution which is kind of bloaty.

src/RFC-0306_DANTemplateRegistration.md
```diff
@@ -0,0 +1,171 @@
...
+the [validator node] SHOULD make the template available to new committee members. New committee members SHOULD confirm the checksum
+with a 2/3 majority of the validator committee to ensure that the correct copy is received.
+
+## Spending the Template Registration UTXO
+
+These cases apply to spending the template registration [UTXO]:
+1. The author MAY spend into another template registration [UTXO]
+* This effectively withdraws (yanks) the previous version of the template.
+* The template name and author SHOULD be identical. This MAY be enforced by covenant.
+* The template version SHOULD be incremented.
```
CjS77, line 150:
```suggestion
* The template version MUST be incremented.
```

src/RFC-0306_DANTemplateRegistration.md
```diff
@@ -0,0 +1,171 @@
...
+* Any live contracts SHOULD upgrade their definitions to run the new template.
```
CjS77, line 151:
```suggestion
* It is RECOMMENDED that live contracts upgrade their definitions to run the new template.
```

CjS77 (commented):
2 minor suggestions

CjS77 (approved):
LGTM

PR #3, opened 2022-07-05
docs: Merge the constitution and the definition UTXOs.

SWvheerden:
Description
---
This creates a proposal to merge the ContractDefinition and ContractConstitution UTXOs into a single Constitution UTXOs governing all of the contract. 

Motivation and Context
---
Currently the RFC states we have two UTXOs, one for defining the contract and one for controlling how its governed. This is a proposal to merge the two into a single UTXO that lives with and can be updated with the contract life of the asset. 

This requires that all information in the ContractDefinition be stored with the ContractConstitution. This is still less data that lives on-chain as we don't require the kernels of two UTXOs, but the actual data contained in them is still the same. 

The biggest requirement of the ContractDefinition UTXO is to identify the contract and its `contract_id`. By merging the two we need an extra base layer consensus rule to ensure that the `contract_id` and its construction fields remain the same across constitution changes. 

The new `contract_id` is defined as: `H(contract_name || asset issuer record)`.  `asset issuer record` is just using a field that is useful for recording purposes as random entropy so we can have duplicate `contract_names`. 

The advantage of going this route is that we have fewer UTXOs needed to manage and run a contract. Although the current RFC is still out of date and states that the ContractDefinition UTXO is spendable, this has been changed to make it static. In practice this makes that the MMR of the entire blockchain will not be prunable after the first ContractDefinition UTXO has been mined. The UTXO set will still be prunable, but not the MMR.

### DIFF
--- a/src/RFC-0312_DANHighLevelSpecification.md
+++ b/src/RFC-0312_DANHighLevelSpecification.md
@@ -234,20 +234,19 @@ assets. This contract can be very simple or highly complex.
 
 The lifecycle of a contract proceeds via these steps:
 
-1. The asset issuer publishes a [contract definition transaction].
-2. The asset issuer publishes a [contract constitution] transaction.
-3. Once this transaction is published, we enter the [acceptance period].
-4. Each validator node that will be managing the contract publishes a [contract acceptance transaction]. The group of
+1. The asset issuer publishes a [contract constitution] transaction.
+2. Once this transaction is published, we enter the [acceptance period].
+3. Each validator node that will be managing the contract publishes a [contract acceptance transaction]. The group of
    validator nodes that manages the contract is called the Validator Node Committee (VNC).
-5. Once the [acceptance period] has expired, the [side-chain initialization period] begins.
-6. The VNC jointly publishes a [side-chain initialization] transaction.
-7. At this point, the contract is considered live, and users can safely interact with the contract on the side-chain.
+4. Once the [acceptance period] has expired, the [side-chain initialization period] begins.
+5. The VNC jointly publishes a [side-chain initialization] transaction.
+6. At this point, the contract is considered live, and users can safely interact with the contract on the side-chain.
    Technically, users do not have to wait until this point. The VNC COULD start processing transactions
    _optimistically_ as soon as the constitution is published, and print the zero-th and first checkpoints once they are
    mined on the base layer. However, this is not generally recommended.
-8. The VNC periodically publishes a [checkpoint] transaction.
-9. Failure to do so can lead to the contract being [abandoned].
-10. The VNC MAY shut the contract down by publishing a [dissolution] transaction.
+7. The VNC periodically publishes a [checkpoint] transaction.
+8. Failure to do so can lead to the contract being [abandoned].
+9.  The VNC MAY shut the contract down by publishing a [dissolution] transaction.
 
 The following sections will discuss each of these steps in more detail.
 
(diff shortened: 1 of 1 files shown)


src/RFC-0312_DANHighLevelSpecification.md
```diff
@@ -293,17 +292,46 @@ requires us to modify the
...
+    * The contract name;
+    * the asset issuer record;
+    * the contract id;
+    * the contract definition;
+    * It MUST include a list of public keys of the proposed VNC;
+    * It MUST include an expiry timestamp before which all VNs must sign and agree to these terms (the [acceptance period]);
+    * It MAY include quorum conditions for acceptance of this proposal (default to 100% of VN signatures required);
+    * If the conditions will unequivocally pass, the acceptance period MAY be shortcut.
+    * There MAY be an initial reward that is paid to the VN committee when the UTXO is spent. This reward is simply included
+        in the value of the `ContractConstitution` UTXO.
```
brianp, line 305: As per the Thursday technical discussion on July 30th, we've removed the initial reward. VN's will not receive an initial reward and we simplify the UTXO needs.

src/RFC-0312_DANHighLevelSpecification.md
```diff
@@ -293,17 +292,46 @@ requires us to modify the
...
+    * The UTXO MUST only be spendable by a multisig of the quorum of VNs performing [side-chain] initialisation. (e.g. a 3
```
brianp, line 306: Additionally the emergency keys, and asset owner can spend checkpoints under specific conditions.

brianp (commented):
I think originally we had planned to put the script inside the contract, and eventually moved that out to the constitution as well to make it easier to update. Which really left nothing by the contract id in the contract definition. So I think now having this all as a single transaction, and hashing the id based on specific fields from it makes sense to me. I don't see a need to require both the definition, and constitution when one would do.

brianp:
@sdbondi > do away with the constitution naming

I think its purpose was sugar for human readability, and we wanted it static overtime to reduce possible confusion.

CjS77:
I vote to keep them separate. It's not an oversight that the current RFC is written this way.

The contract (what code is being run) and the constitution (how the code is managed) are disparate enough concepts that they warrant separate entries on the blockchain.

We want to keep a queryable index of live contracts, and a queryable index of constitutions. Keeping these separate simplifies the database management. The cost in terms of storage space is worth the trade-off IMHO. 

Moreover, the contract definition very seldom changes, if ever. It should be treated as near-immutable. If this UTXO is spent, observers know that the code governing the contract (i.e. the contract itself) has likely changed, and this can and should fire off all sorts of red flags.

I'd argue that constitutions will change an order of magnitude more frequently than the contract, so merging the two means that observers must do many more checks on the contract code than they otherwise would need to.

SWvheerden:
> I vote to keep them separate. It's not an oversight that the current RFC is written this way.
> 
> The contract (what code is being run) and the constitution (how the code is managed) are disparate enough concepts that they warrant separate entries on the blockchain.
> 
> We want to keep a queryable index of live contracts, and a queryable index of constitutions. Keeping these separate simplifies the database management. The cost in terms of storage space is worth the trade-off IMHO.
> 
> Moreover, the contract definition very seldom changes, if ever. It should be treated as near-immutable. If this UTXO is spent, observers know that the code governing the contract (i.e. the contract itself) has likely changed, and this can and should fire off all sorts of red flags.
> 
> I'd argue that constitutions will change an order of magnitude more frequently than the contract, so merging the two means that observers must do many more checks on the contract code than they otherwise would need to.


What do we do with versioning of code, do we not lock that in as well? That will almost definitely update. Then we need to update this. 

The only reason I am against them separately is the MMR prunability. 


Here is the current rfc for the templates:
https://github.com/tari-project/rfcs/pull/2

CjS77:
Yes, and a code version update is
1. exactly what people will be looking to audit
2. happen much less frequently than constitution changes for the vast majority of contracts

CjS77:
The asset issuer basically ceases to be relevant after the contract initialisation. 

If the AI does not add themselves to the CC they basically relinquish any and all control of the contract.

CjS77:
I see no point in spending effort to resolve conflicts when this RFC will be deprecated soon in favour of a better proposal being developed by the community, so just going to close it.

PR #4, opened 2022-07-06
Abandonment proposal

SWvheerden:
Description
---
This updates the contract abandonment process to the current to-be-implemented code.


brianp (commented):
This makes sense to me. It's giving a name to an observable state. It doesn't functionally change _much_ but helps with discussions and creates clearer stages in the cycle.

SWvheerden:
The only reason for adding `chain monitoring for CC proposals` is to stop a case where you have backup keys and CC trying to both take control of an asset at the same time. 

And this is more in line with making sure there are no gaps in the security model anywhere. If I have learned anything from all the attacks and hacks from Eth/Solana etc, is that if there is a possible gap, it's a matter of time until someone takes advantage of it and exploits it.

CjS77 (approved):
We can merge this to be part of the historical record.

PR #8, opened 2022-07-18
Submarine swaps

SWvheerden:
Discussion of how submarine swaps can work the second layer


CjS77 (approved):
Can merge this into proposal as is, and further discussions can be made via additional PRs

PR #10, opened 2022-07-22
feat: rfc for burning

SWvheerden:
Description
---
This will us to do burns on UTXOs. 

Motivation and Context
---
Many second layer functionality requires the burning of main chain coins. This will enable Tari to be burned and be removed from the emission.

### DIFF
--- a/src/RFC-0122_Burning.md
+++ b/src/RFC-0122_Burning.md
@@ -1,6 +1,6 @@
-# RFC-0122/Burning
+# RFC-0260/Burning
 
-## Base Layer Consensus
+## Consensus-level burn transactions
 
 ![status: draft](theme/images/status-draft.svg)
 
@@ -62,8 +62,8 @@ to denote burned coins. Using [RFC-0201: TariScript](RFC-0201_TariScript.md), Ta
 remove the coins from circulation. This RFC details a method to remove coins from circulation permanently.
 
 An excellent example of using burned coins is for a [Perpetual One-way Peg](https://medium.com/@RubenSomsen/21-million-bitcoins-to-rule-all-sidechains-the-perpetual-one-way-peg-96cb2f8ac302) this allows users 
-create utility tokens on a sidechain or in Tari's case the Dan. By not allowing funds to be moved back, the sidechain gets a total utilitarian value
-. All value speculation will be removed as the sidechain will only ever be as valuable as the main coin.
+create utility tokens on a sidechain or in Tari's case the DAN. By not allowing funds to be moved back, the sidechain gains an auditable total utilitarian value
+. Value speculation is largely removed as the sidechain will only ever be as valuable as value of coins burnet in the peg.
 
 
 ## Introduction
@@ -117,7 +117,7 @@ pub struct TransactionKernel {
 
 }
 ```
-Each kernel also has a field called [KernelFeatures](https://github.com/tari-project/tari/blob/2ca06724f0ab7c63eb0b6caab563372f353f4348/base_layer/core/src/transactions/transaction_components/kernel_features.rs#L36) that unique properties of the kernel. These fields need to be expanded to include a burn type.
+Each kernel also has a field called [KernelFeatures](https://github.com/tari-project/tari/blob/2ca06724f0ab7c63eb0b6caab563372f353f4348/base_layer/core/src/transactions/transaction_components/kernel_features.rs#L36) that defines the properties of the kernel. The feature set needs to be expanded to include a burn type.
 Inside this field, we track every possible type of output. We need to add another type here called `BURNED_KERNEL`. 
 
 ```rust,ignore
@@ -137,11 +137,12 @@ This stops a node from publishing an aggregated inflated commitment in the kerne
 ### Consensus rules changes
 
 When a block or transaction is received with a burned output, there:
-* MUST be a kernel containing a commitment matching exactly one and only one output that is flagged as burned.
+* the number of `BURNED` outputs MUST equal the number of `BURNED_KERNEL` kernels exactly,
+* the commitment values of each burnt output MUST match the commitment value of each corresponding `BURNED_KERNEL` exactly.
 
 ### Chain balance equation changes 
 
-When checking the total chain emission, the following equation must hold:
+Currently, when checking the total chain emission, the following equation must hold:
 $$
 \begin{align}
 &\sum_i\mathrm{C_{i}} \stackrel{?}{=} E_l \cdot H + \sum_k\mathrm{K_k} + \text{offset}\\\\
@@ -153,7 +154,11 @@ $$
 \tag{1}
 $$
 
+To include burnt outputs in the balance, we can immediately mark burnt outputs as spent (and thus allow them to be pruned), and introduce the additional requirement of tracking the total sum of all burnt output commitments.
 
+This sum is easily verified by summing all `BURNED_KERNEL` commitments in the blockchain history. These are available to all nodes since the kernels are never pruned.
+
+The balance equation then becomes
 $$
 \begin{align}
 &\sum_i\mathrm{C_{i}} \stackrel{?}{=} E \cdot H + \sum_k\mathrm{K_k} + \text{offset} - \sum_m\mathrm{Cburn_{m}}\\\\
--- a/src/RFC-0122_Burning.md
+++ b/src/RFC-0122_Burning.md
@@ -61,9 +61,9 @@ Blockchains have used the burn method to destroy coins in circulation and block 
 to denote burned coins. Using [RFC-0201: TariScript](RFC-0201_TariScript.md), Tari can also use this method, but this does not explicitly
 remove the coins from circulation. This RFC details a method to remove coins from circulation permanently.
 
-A very good example of using burned coins is for a [Perpetual One-way Peg](https://medium.com/@RubenSomsen/21-million-bitcoins-to-rule-all-sidechains-the-perpetual-one-way-peg-96cb2f8ac302) this allows users 
+An excellent example of using burned coins is for a [Perpetual One-way Peg](https://medium.com/@RubenSomsen/21-million-bitcoins-to-rule-all-sidechains-the-perpetual-one-way-peg-96cb2f8ac302) this allows users 
 create utility tokens on a sidechain or in Tari's case the Dan. By not allowing funds to be moved back, the sidechain gets a total utilitarian value
-and all value speculation will be removed as the sidechain will only ever be as valuable as the main coin.
+. All value speculation will be removed as the sidechain will only ever be as valuable as the main coin.
 
 
 ## Introduction
@@ -71,9 +71,13 @@ and all value speculation will be removed as the sidechain will only ever be as 
 To completely remove coins from circulation, we need to mark outputs as burned, and we need to change the balance equation to
 allow pruned nodes and non pruned nodes to still verify the integrity and emission of Tari.
 
+### TL;DR
+In order to get burns working, we flag each desired output as burned with a flag. To calculate the emission of the network, we need
+to store the commitment of the burned output in a kernel. This will allow any pruned node to verify the emission of the network and all burned outputs. 
+
 ### Transaction output changes
 
-Currently, each transaction output has a field called [OutputFeatures](https://github.com/tari-project/tari/blob/f7f913c873c9f5d373f99149e52c26a0dd32f03f/base_layer/core/src/transactions/transaction_components/output_features.rs#L62) that tracks the unique properties of each output. 
+Each transaction output has a field called [OutputFeatures](https://github.com/tari-project/tari/blob/f7f913c873c9f5d373f99149e52c26a0dd32f03f/base_layer/core/src/transactions/transaction_components/output_features.rs#L62) that tracks the unique properties of each output. 
 Inside this field, we track every possible type of output. We need to add another type here called burned. 
 
 ```rust,ignore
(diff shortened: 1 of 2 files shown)


src/RFC-0122_Burning.md
```diff
@@ -0,0 +1,193 @@
...
+& \text{for each output}, i, \\\\
+& \text{for each input}, j, \\\\
+& \text{for each kernel excess}, k \\\\
+& \textit{offset }\text{is the total kernel offset} \\\\
+& \text{and }V_l \text{ is the block reward for block } l\\\\
+\end{align}
+\tag{2}
+$$
+
+This equation need to be modified to allow checking the burned number as well, so we change it to:
```
stringhandler, line 133:
```suggestion
This equation needs to be modified to allow checking the burned number as well, so we change it to:
```

src/RFC-0122_Burning.md
```diff
@@ -0,0 +1,193 @@
...
+\begin{align}
+burned\_total = \sum_j\mathrm{C_{j}}  \\\\
+\text{for burned output}, i, \\\\
+\end{align}
+\tag{1}
+$$
+
+### Block propagation consensus rules changes
+
+When a block is received, a base node will typically check the following equation to ensure that the emission is correct:
```
stringhandler, line 120: Thi Latex doesn't seem to render in Github's markdown, but it might be fine in mdbook

src/RFC-0122_Burning.md
```diff
@@ -0,0 +1,193 @@
...
+& \textit{offset }\text{is the total kernel offset} \\\\
+& \text{and }V_l \text{ is the block reward for block } l\\\\
+\end{align}
+\tag{3}
+$$
+
+We make the following additions to the consensus rules that block propagation must be checked.
+
+When a node receives a block with the node:
+* MUST for each output marked as burned, flag that output as spent immediately upon inserting it as an unspent output.
```
stringhandler, line 150:
```suggestion
* MUST for each output marked as burned, flag that output as spent immediately upon inserting it as an output. 
```

stringhandler (commented):
I would also add to the header:
A merkle mountain range root and length of all the commitments in the burnt set, so that another chain can create a merkle proof that a commitment has been burnt on Tari. The length of the MMR is added for the same reason as the kernel and output lengths.

src/RFC-0122_Burning.md
SWvheerden: Yeah github does not render correctly

src/RFC-0122_Burning.md
```diff
@@ -115,6 +115,27 @@ burned\_total = \sum_j\mathrm{C_{j}}  \\\\
 \tag{1}
 $$
 
+### Creation of burned output
+
+When creating the output, we need to be able to generate a Proof of Burn for the burned output. 
+In the header, we already commit to each and every [UTXO] ever created. There already is a proof of the existence
+for this burned output, but this proof is only to the complete output and not just the commitment.
```
stringhandler, line 122:
```suggestion
for this burned output, but this proof is only for the complete output and not just the commitment. 
```

src/RFC-0122_Burning.md
```diff
@@ -115,6 +115,27 @@ burned\_total = \sum_j\mathrm{C_{j}}  \\\\
 \tag{1}
 $$
 
+### Creation of burned output
+
+When creating the output, we need to be able to generate a Proof of Burn for the burned output. 
+In the header, we already commit to each and every [UTXO] ever created. There already is a proof of the existence
+for this burned output, but this proof is only to the complete output and not just the commitment. 
+
+To get around this, we enforce default values for most of the output except the version and commitment field.
```
stringhandler, line 124: I don't like this approach. Feels very hacky

src/RFC-0122_Burning.md
```diff
@@ -115,6 +115,27 @@ burned\_total = \sum_j\mathrm{C_{j}}  \\\\
...
 
+### Creation of burned output
+
+When creating the output, we need to be able to generate a Proof of Burn for the burned output. 
+In the header, we already commit to each and every [UTXO] ever created. There already is a proof of the existence
+for this burned output, but this proof is only to the complete output and not just the commitment. 
+
+To get around this, we enforce default values for most of the output except the version and commitment field. 
+
+And the output, when burned, MUST have the following field values
```
stringhandler, line 126: I don't like these requirements. It is limiting because other chains might also want to use this functionality, and they might need a specific metadata, or script

src/RFC-0122_Burning.md
```diff
@@ -181,6 +202,46 @@ $$
 
 When verifying total chain emission, equation (5) MUST hold.
 
+
+### Option 2 for Proof of burn
+
+If we don't want to look into the output MMR, we can include a second MMR root and size counter to the header to track the commitment.
```
stringhandler, line 208:
```suggestion
If we don't want to look into the output MMR, we can add a second MMR root and size counter to the header to track the commitment. 
```

src/RFC-0122_Burning.md
```diff
@@ -181,6 +202,46 @@ $$
...
+}
+```
+
+[burned_mr]: #burned_mr "Burned Merkle root"
+
+This is the Merkle root of the burned outputs.
+
+The kernel_mr MUST conform to the following:
+
+* Must be transmitted as an array of unsigned 8-bit integers (bytes) in little-endian format.
```
stringhandler, line 238: Rather add this info to the other relevant hashing RFC. 

It should also be hashed as a VarInt. 

Transfer doesn't really matter, since it is sent via Protobuf

stringhandler (commented):
I much prefer option 2

src/RFC-0122_Burning.md
SWvheerden: I want us to enforce these rules anyway, regardless of the route we go. I don't think there are any security gaps around allowing scripts etc on the burned outputs, but I don't think we should take that chance. As an additional benefit, this is smaller per output, if all the fields are essentially blank.

src/RFC-0122_Burning.md
SWvheerden: We should probably update the other consensus rfc then for the other hashes.

SWvheerden:
> I much prefer option 2

I am about 90% for option 1. 
I get why we would want an "easier" proof, but with implementation, I highly suspect that when implementing this we will use the existing methods and just ask for the utxo. I think option 2 will be an unused feature + data on the header. Therefore I propose we use method 1 when if/when we do want to do something like this.

src/RFC-0122_Burning.md
stringhandler: The space doesn't really make a difference, since these are pruned immediately. I'm not convinced these need to be blank

stringhandler:
Remember that outputs will need to be hashed according to versions. The hashed fields will not always be the same. That means a client must remember the entire output, or at a minimum, the commitment and the version, to generate the proof.

In addition, clients outside of the Tari ecosystem where these funds may need to be claimed now need to implement every hash version implemented in Tari, and they must implement it exactly correctly, otherwise they cannot verify the proof. It is much simpler for them to do a simple hash and Merkle proof of the commitment only

src/RFC-0122_Burning.md
```diff
@@ -0,0 +1,252 @@
...
+## Related Requests for Comment
+
+* [RFC-0100: Base Layer](RFC-0100_BaseLayer.md)
+* [RFC-0120: Consensus](RFC-0120_Consensus.md)
+
+## Description
+
+Blockchains have used the burn method to destroy coins in circulation and block their use forever. Most chains use a reclaimable address
+to denote burned coins. Using [RFC-0201: TariScript](RFC-0201_TariScript.md), Tari can also use this method, but this does not explicitly
+remove the coins from circulation. This RFC details a method to remove coins from circulation permanently.
```
CjS77, line 62: Add a paragraph explaining the motivation for adding this.(i.e. it serves as a Perpetual 1-way Peg for redeeming utility tokens on the DAN)

A link to [P1wP](https://medium.com/@RubenSomsen/21-million-bitcoins-to-rule-all-sidechains-the-perpetual-one-way-peg-96cb2f8ac302) as precedent would be good context too.

src/RFC-0122_Burning.md
```diff
@@ -0,0 +1,252 @@
...
+
+
+## Introduction
+
+To completely remove coins from circulation, we need to mark outputs as burned, and we need to change the balance equation to
+allow pruned nodes and non pruned nodes to still verify the integrity and emission of Tari.
+
+### Transaction output changes
+
+Currently, each transaction output has a field called [OutputFeatures] that tracks the unique properties of each output.
```
CjS77, line 72: Link is broken

src/RFC-0122_Burning.md
CjS77: hmm.. for example, if you want to indicate on the second layer which side-chain the burned funds should be redeemed on; or provide a proof of ownership certificate in a script that can be picked up on the 2nd layer.

I think there are several use cases that we'd like to leverage with the metadata and script fields.

src/RFC-0122_Burning.md
```diff
@@ -0,0 +1,252 @@
...
+& \textit{offset }\text{is the total kernel offset} \\\\
+& \text{and }V_l \text{ is the block reward for block } l\\\\
+\end{align}
+\tag{3}
+$$
+
+We make the following additions to the consensus rules that block propagation must be checked.
+
+When a node receives a block with the node:
+* MUST for each output marked as burned, flag that output as spent immediately upon inserting it as an output.
```
CjS77, line 171: This is implicitly taken care of in the overall balance (5). As it's an implementation detail, it's not a MUST do.

src/RFC-0122_Burning.md
```diff
@@ -0,0 +1,252 @@
...
+pub struct BlockHeader {
+    /// Version of the block
+    pub version: u16,
+    ...
+    /// Sum of kernel offsets for all kernels in this block.
+    pub total_kernel_offset: BlindingFactor,
+    /// Sum of script offsets for all kernels in this block.
+    pub total_script_offset: BlindingFactor,
+    /// Sum of all burned output commitments in this block.
+    pub burned_total: Commitment,
```
CjS77, line 100: I would actually propose making this the sum of burnt outputs in this block PLUS the value from the previous header (or ZERO if it's the genesis block). i.e. a running total of all BTXOs

src/RFC-0122_Burning.md
```diff
@@ -0,0 +1,252 @@
...
+    /// Nonce increment used to mine this block.
+    pub nonce: u64,
+    /// Proof of work summary
+    pub pow: ProofOfWork,
+}
+```
+
+This field must contain the sum of all the commitments of all outputs in the block marked as spent.
```
CjS77, line 108:
```suggestion
This field must contain the sum of all the commitments of all outputs in the block marked as BURNt. 
```

src/RFC-0122_Burning.md
CjS77: Surely the only change from the _status quo_ is that we must commit to the BURN flag somewhere in an MMR so it can never be unburnt. This in addition to a consensus rule forbidding BTXOs to be used as inputs, of course.

CjS77 (commented):
Some details to be fleshed out, but overall, looking good.

src/RFC-0122_Burning.md
stringhandler: I don't really have a preference here, so happy to go with that as well

SWvheerden: Any reason for this?
I thought about doing it that way but decided against it due to the fact of how we approach and validate blocks in code currently. 
When validating a block, we don't look at overall balance, we only do validation on the block. 
for example:
Is the reward correct for this block,
does this block balance, etc. 
We never do chain balance validation etc, we make the assumption that if the currently stored state is correct we only have to validate the delta and not the entire state again. We do the same with the offsets. This follows that mindset to only validate the delta.

stringhandler: We do `parent = prev.hash` and `height = prev.height + 1`, as well as checking MMR roots so I don't think there's much extra work in validating the running total vs total for the block.

src/RFC-0122_Burning.md
SWvheerden: mmm, okay, I have been convinced that we should not enforce anything here.

src/RFC-0122_Burning.md
SWvheerden: This has an effect on the MMR of the outputs. 
If an output is marked as spent, then that reflects on the bitmap of all spent outputs. And this gets committed to in the MMR of the outputs.

src/RFC-0122_Burning.md
SWvheerden: Don't get me wrong, it's definitely doable, 
This way just fits in better with the current validation and header checks. 
Hence why I asked if there is specifically a reason why we want to do it this way?

CjS77:
Another suggestion: `BURN` commitments should not be BTXOs at all.

Any transaction with a burn can must the commitment in the _kernel_ and cover it by the kernel signature.

Pros:
* It's signed, so pruned nodes will be able to validate the sum of burnt commitments. 
* Kernels are never pruned, so no extra logic required to maintain an accurate sum
* No changes to current output logic required.

Cons:
* Burns are associated with a kernel, which is a small privacy hit.

src/RFC-0122_Burning.md
```diff
@@ -0,0 +1,180 @@
...
+
+## Description
+
+Blockchains have used the burn method to destroy coins in circulation and block their use forever. Most chains use a reclaimable address
+to denote burned coins. Using [RFC-0201: TariScript](RFC-0201_TariScript.md), Tari can also use this method, but this does not explicitly
+remove the coins from circulation. This RFC details a method to remove coins from circulation permanently.
+
+An excellent example of using burned coins is for a [Perpetual One-way Peg](https://medium.com/@RubenSomsen/21-million-bitcoins-to-rule-all-sidechains-the-perpetual-one-way-peg-96cb2f8ac302) this allows users 
+create utility tokens on a sidechain or in Tari's case the Dan. By not allowing funds to be moved back, the sidechain gets a total utilitarian value
+. All value speculation will be removed as the sidechain will only ever be as valuable as the main coin.
```
CjS77, line 66:
```suggestion
. Value speculation is largely removed as the sidechain will only ever be as valuable as value of coins burnet in the peg.
```

src/RFC-0122_Burning.md
```diff
@@ -0,0 +1,180 @@
...
+    pub excess: Commitment,
+    /// An aggregated signature of the metadata in this kernel, signed by the individual excess values and the offset
+    /// excess of the sender.
+    pub excess_sig: Signature,
+    /// This is an optional field that must be set if the transaction contains a burned output.
+    pub burn_commitment: Option<Commitment>,
+
+}
+```
+Each kernel also has a field called [KernelFeatures](https://github.com/tari-project/tari/blob/2ca06724f0ab7c63eb0b6caab563372f353f4348/base_layer/core/src/transactions/transaction_components/kernel_features.rs#L36) that unique properties of the kernel. These fields need to be expanded to include a burn type.
```
CjS77, line 120:
```suggestion
Each kernel also has a field called [KernelFeatures](https://github.com/tari-project/tari/blob/2ca06724f0ab7c63eb0b6caab563372f353f4348/base_layer/core/src/transactions/transaction_components/kernel_features.rs#L36) that defines the properties of the kernel. The feature set needs to be expanded to include a burn type.
```

src/RFC-0122_Burning.md
```diff
@@ -0,0 +1,180 @@
...
+
+For each burned output in a transaction, there needs to be a kernel. This means that if a transaction has two burned outputs, it needs at least two kernels. 
+This stops a node from publishing an aggregated inflated commitment in the kernel. 
+
+
+
+### Consensus rules changes
+
+When a block or transaction is received with a burned output, there:
+* MUST be a kernel containing a commitment matching exactly one and only one output that is flagged as burned.
```
CjS77, line 140:
```suggestion
* the number of `BURNED` outputs MUST equal the number of `BURNED_KERNEL` kernels exactly,
* the commitment values of each burnt output MUST match the commitment value of each corresponding `BURNED_KERNEL` exactly.
```

src/RFC-0122_Burning.md
```diff
@@ -0,0 +1,180 @@
...
+
+### Chain balance equation changes 
+
+When checking the total chain emission, the following equation must hold:
```
CjS77, line 144:
```suggestion
Currently, when checking the total chain emission, the following equation must hold:
```

src/RFC-0122_Burning.md
```diff
@@ -0,0 +1,180 @@
...
+&\sum_i\mathrm{C_{i}} \stackrel{?}{=} E_l \cdot H + \sum_k\mathrm{K_k} + \text{offset}\\\\
+& \text{for each unspent output}, i, \\\\
+& \text{for each kernel excess}, k \\\\
+& \textit{offset }\text{is the total kernel offset} \\\\
+& \text{and }E_l\text{ is total emission up to block } l\\\\
+\end{align}
+\tag{1}
+$$
+
+
```
CjS77, line 156:
```suggestion
To include burnt outputs in the balance, we can immediately mark burnt outputs as spent (and thus allow them to be pruned), and introduce the additional requirement of tracking the total sum of all burnt output commitments.

This sum is easily verified by summing all `BURNED_KERNEL` commitments in the blockchain history. These are available to all nodes since the kernels are never pruned.

The balance equation then becomes
```

stringhandler (approved):

PR #11, opened 2022-09-08
docs: update emoji ID

AaronFeickert:
Description
---
[RFC-0152](https://rfc.tari.com/RFC-0152_EmojiId.html), which defines emoji IDs, is outdated. This brings it up to date and includes changes from [PR 4639](https://github.com/tari-project/tari/pull/4639), which updates the checksum algorithm.

Motivation and Context
---
The following are outdated in RFC-0152:
- Versioning is not used in the codebase.
- The emoji character set in the codebase contains only 256 symbols.
- The encoding method in the RFC is not used in the codebase.

Further, it is likely that the checksum algorithm will change with PR 4639.

### DIFF
--- a/src/RFC-0152_EmojiId.md
+++ b/src/RFC-0152_EmojiId.md
@@ -2,7 +2,7 @@
 
 ## Emoji Id specification
 
-![Status: Outdated](theme/images/status-outofdate.svg)
+![status: draft](theme/images/status-draft.svg)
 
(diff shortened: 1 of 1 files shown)


AaronFeickert:
This is blocked by [PR 4639](https://github.com/tari-project/tari/pull/4639).

CjS77 (approved):
Thanks for the update!

PR #14, opened 2022-09-27
feat: add aggregate signatures to transaction inputs and outputs

hansieodendaal:
Description
---
Added aggregate metadata and script signatures to the transaction output and input (RFC-0201) and multi-sig opcodes (RFC-0202).

Motivation and Context
---
This is handy for multiple parties (`m-of-n`) to create outputs and spend outputs. 

To create a multi-party output `n-of-n` receiver parties need to:
- agree on the `m-of-n` TariScript script out of band;
- create an aggregated receiver portion of the metadata signature if the spending key is sharded. 

To spend an `m-of-n` input, `m` parties need to agree on the `m-of-n` TariScript input out of band and create an aggregated script signature.

_**Main additions to RFC-0201 text is shown below**_

---
image

image

image

image

---

How Has This Been Tested?
---
`mdbook serve` renders fine

### DIFF
--- a/src/RFC-0201_TariScript.md
+++ b/src/RFC-0201_TariScript.md
@@ -406,9 +406,6 @@ Note that:
   metadata signature will be invalidated.
 - We provide the complete script on the output.
 
-_(See [Multi-party transaction output](#multi-party-transaction-output) for a discussion ot the multi-party version of a 
-transaction output.)_
-
 ### Transaction input changes
 
 The current definition of an input is
@@ -486,9 +483,6 @@ but obtained by executing the script with the provided input data. Because this 
 private key \\(k\_{Si}\\), it ensures that only the owner can provide the input data \\(\input_i\\) to the 
 TransactionInput. 
 
-_(See [Multi-party transaction input](#multi-party-transaction-input) for a discussion ot the multi-party version of a
-transaction output.)_
-
 ### Script Offset
 
 For every transaction an accompanying [script offset] \\( \so \\) needs to be provided. This is there to prove that every  
@@ -502,9 +496,6 @@ $$
 \end{aligned}
 \tag{16}
 $$
-
-_(See [Multi-party script offset](#multi-party-script-offset) for a discussion ot the multi-party version of the
-script offset.)_
 
 Verification of (16) will entail:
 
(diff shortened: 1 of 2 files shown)


SWvheerden (dismissed):
This looks good, although this will clash with some of #15 which adds in more complete descriptions to the multisig opcodes.

CjS77:
This is interesting, but it is should be in its own RFC.

* If this is a discussion of an application of the currently implemented protocol (which it looks like it is, largely), then it should go in the Discussions chapter (maybe we should also create an Applications section). The main reason for this ask is to keep the original RFC focused (it's already very technical and difficult to follow).
* Post-mainnet, the new opcodes would definitely need a new RFC. Pre-mainnet, if there's broad consensus to implement before mainnet, it can go into the current RFC. Otherwise it should also be separate.

hansieodendaal:
> This is interesting, but it is should be in its own RFC.
> 

Thanks for the review, but I do not think I agree with this request, as the multi-party considerations section cannot be read on its own or without continuously referring back to RFC-0201, making it haphazard. The main theme of this PR is the impact and changes to the metadata and script signatures.

CjS77:
> as the multi-party considerations section cannot be read on its own or without continuously referring back to RFC-0201, making it haphazard.

This is not true. It not haphazard to bring in applications after the basics -- you don't teach Calculus when you're introducing algebra based on the argument that calculus uses Algebra.

The key point I was making is that a newcomer reading this for the first time is already dealing with some novel concepts just to grasp TariScript. The reference to multiparty txs throughout the text makes this completely overwhelming.

The other problem is that the MP discussion is more about secret sharing than it is about TariScript.

tl;dr I still favour a separate RFC dedicated to applications of multiparty transactions. As a compromise, just remove references to multiparty txs in the main text. 

Having reviewed the layout of this RFC, I see there are a few sections that _are_ core to TariScript that have been buried in the Examples section that need to be moved up (ex cut-through and security); but we can do that in another PR.

hansieodendaal:
> tl;dr I still favour a separate RFC dedicated to applications of multiparty transactions. As a compromise, just remove references to multiparty txs in the main text.

Accepted, thanks

stringhandler:
I am going to merge this PR, but then split it out into a new RFC

stringhandler (approved):

PR #19, opened 2022-09-30
fix(rfc201): move cut through explanation earlier

stringhandler:
Although cut-through prevention is mentioned early in the document, the reasoning was included in the explanation, which was very far down the document. I've moved the explanation earlier, but kept the implementation details where they were.


hansieodendaal (approved):
LGTM

SWvheerden (approved):
LGTM

PR #33, opened 2022-10-20
fix: clean up organisation of rfcs

CjS77:
The main idea behind this PR is
- only specifications directly related to the implementation of the base layer and DAN go under the base layer and DAN chapters respectively.
- Applications of the L1 or L2, thought experiments and other proposals are moved into the "Proposals" chapter.

This lets them stay in "draft", perhaps indefinitely, without making the L1 and L2 look 'unfinished' -- since the proposals are not integral to the core protocols.

But these ideas and proposals still belong in the broader RFC doc, so creating a new chapter for them to live makes the most sense.

A few RFCs have also been moved into the proposals chapter as part of this PR, meaning that the L1 and DAN chapters should now only contain RFCs that are integral to the functioning of those protocols, and MUST be moved to STABLE status ahead of mainnet.


stringhandler (approved):
ACK

PR #34, opened 2022-10-26
feat: stabilise rfc-0131

CjS77:
Description
---

* A few small edits - content is materially unchanged.
* Add change log entry
* Change status to STABLE

### DIFF
--- a/src/RFC-0131_Mining.md
+++ b/src/RFC-0131_Mining.md
@@ -2,7 +2,7 @@
 
 ## Full-node Mining on Tari Base Layer
 
-![status: draft](theme/images/status-draft.svg)
+![status: stable](theme/images/status-stable.svg)
 
 **Maintainer(s)**: [Hansie Odendaal](https://github.com/hansieodendaal), [Philip Robinson](https://github.com/philipr-za)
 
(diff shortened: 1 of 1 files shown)


CjS77:
ACK

sdbondi (approved):

sdbondi:
ACK

SWvheerden:
ACK

src/RFC-0131_Mining.md
```diff
@@ -135,6 +134,9 @@ The hash MUST be constructed with the following encoded with [consensus encoding
 - total kernel offset
 - total script offset
 
+This hash is used in both the SHA-3 and RandomX proof-of-work algorithms. The header version for the Tari Genesis 
+block is 1.
```
SWvheerden, line 138: nit: We currently count from 0, 
Either we need to update this to 0, or we need to update the code to count the header version from 1.

SWvheerden (commented):
Looks good, we just need to decide what number we want the block version to run from for gen

src/RFC-0131_Mining.md
CjS77: The code already has the current version at 1.
Version 0 was for Dibbler.

SWvheerden: we had 0 for esme as well. 
https://github.com/tari-project/tari/blob/6ff1c02d3451d856a7c0c979109aaae99dc38ca1/base_layer/core/src/consensus/consensus_constants.rs#L519

SWvheerden: we changed to 1 when we changed the PoW

CjS77: Ok, but you agree that the current version is 1; and as the code stands, the genesis block will also be version 1.

CjS77: Going to close this convo with the following:
- Since it currently matches the code, it's better to leave as is. Otherwise auditors will flag it.
- If the code changes ahead of mainnet, this RFC must be amended to match.

sdbondi (approved):

PR #36, opened 2022-10-26
feat: stabilise rfc-0173

CjS77:
Description
---

* Stabilises RFC-0173
* Minor edits - content is materially unchanged
* Change log added

### DIFF
--- a/src/RFC-0173_Versioning.md
+++ b/src/RFC-0173_Versioning.md
@@ -75,21 +75,34 @@ delimit when a new message starts in a byte stream and also are used to indicate
 is speaking on, such as TestNet or Mainnet.
 
 Tari message packets are encapsulated using the Noise protocol, so we do not need the delimiting functionality of these 
-bytes but Tari will include a single WireMode byte at the beginning of every connection session. This byte will indicate 
+bytes.
+Once we have a Noise socket we are able to send/receive bytes as with any other socket, but those bytes are
+encrypted over the wire. Using that socket we send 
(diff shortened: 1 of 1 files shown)


CjS77:
ACK

src/RFC-0173_Versioning.md
```diff
@@ -74,7 +74,7 @@ In the Bitcoin P2P protocol messages are preceded by 4
 delimit when a new message starts in a byte stream and also are used to indicate which of the Bitcoin networks the node 
 is speaking on, such as TestNet or Mainnet.
 
-Tari message packets are encapsulated using the Noise protocol so we do not need the delimiting functionality of these 
+Tari message packets are encapsulated using the Noise protocol, so we do not need the delimiting functionality of these
```
sdbondi, line 77: The wiremode description is accurate however I wouldn't say that the reason we don't use a message delimiter is because of the noise socket. Once you have a noise socket you are able to send/receive bytes as with any other socket, but those bytes are encrypted over the wire. Using that socket we send [yamux packets](https://github.com/hashicorp/yamux/blob/master/spec.md) and messaging/rpc messages are [length-delimited](https://docs.rs/tokio-util/latest/tokio_util/codec/length_delimited/).

sdbondi: Maybe we should change the tense to present?

CjS77: Thanks. Updated

sdbondi (approved):

PR #41, opened 2022-11-04
fix(stable base nodes): minor updates and move to stable

brianp:
Description
---
Added a little extra info regarding pruned nodes and peer validation.
Moved RFC to stable.

Motivation and Context
---
Gett'n things updated.

### DIFF
--- a/src/RFC-0110_BaseNodes.md
+++ b/src/RFC-0110_BaseNodes.md
@@ -78,7 +78,7 @@ Tari Base Nodes MUST carry out the following tasks:
 * provide historical block information to peers that are syncing.
 
 Once the Digital Assets Network (DAN) goes live, Base Nodes will also need to support the tasks described in
-[RFC-0300_DAN](RFCD-0300_DAN.md). These requirements may involve but are not limited to:
+[RFC-0303_DAN](RFC-0303_DanOverview.md). These requirements may involve but are not limited to:
 * maintain an index of validator node registrations;
 
 To carry out these tasks effectively, Base Nodes SHOULD:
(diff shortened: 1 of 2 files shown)


src/RFC-0110_BaseNodes.md
```diff
@@ -78,7 +78,8 @@ Tari Base Nodes MUST carry out the following tasks:
 * provide historical block information to peers that are syncing.
 
 Once the Digital Assets Network (DAN) goes live, Base Nodes will also need to support the tasks described in
-[RFC-0300_DAN](RFCD-0300_DAN.md). These requirements are omitted for the moment.
+[RFC-0300_DAN](RFCD-0300_DAN.md). These requirements may involve but are not limited to:
```
CjS77, line 81: RFC-0304

CjS77 (commented):
LG, but RFC link needs to be updated

CjS77:
Points/inaccuracies that should be addressed and/or confirmed before merging:

* Cut-through has been deacivated in Tari
* Update "Block/Transaction Weight" section language to be authoritative rather than the conditional form (and confirm that the values match the code).
* BroadcastStrategy might be an outdated reference.

brianp:
> Points/inaccuracies that should be addressed and/or confirmed before merging:

- ✅ Cut-through has been deactivated in Tari
Confirmed as per RFC-0201_TariScript cut-through has been deactivated in Tari to prevent a bad actor from utilizing cut through to spend a UTXO that they know the blinding factor too, but Tari script is being used enforce a particular public key.
- ✅ Update "Block/Transaction Weight" section language to be authoritative rather than the conditional form (and confirm that the values match the code).
Values were already confirmed, language updated, table formatting fixed.
- ✅ BroadcastStrategy might be an outdated reference.
Still seems relevant but updated the language to be more specific about which strategy as multiple are mentioned in the glossary.

stringhandler (approved):

PR #42, opened 2022-11-04
fix(htlc stabilising): update htlc rfcs

brianp:
Description
---
Originally I deprecated this RFC in favour of moving the content into the TariScript opcodes examples section. The two examples are similar but hidden in the details of the opcodes example is that it's a multiparty transaction where the one in the 0230 example is not. So I've left the RFC separated as it more closely matches the implementation and atomic swap use case.

Updated the naming in the opcode rfc just for disambiguation purposes.

Motivation and Context
---
Stabilising 

How Has This Been Tested?
---


CjS77:
ACK

PR #46, opened 2022-11-10
docs: update rfc_0202 for code review audit

hansieodendaal:
Description
---
Updated RFC-0202 for code review/audit by bringing it in line with the actual code.

Motivation and Context
---
See above

How Has This Been Tested?
---
`mdbook serve`


src/RFC-0202_TariScriptOpcodes.md
```diff
@@ -220,162 +217,172 @@ Reverse rotation. The top stack item moves into 3rd place, e.g. `abc => bca`.
 
 #### GeZero
 
-Pops the top stack element as `val`. If `val` is greater than or equal to zero, push a 1 to the stack, otherwise push 0.
+Pops the top stack element as `val`. If `val` is greater than or equal to zero, push a $1$ to the stack, otherwise
+push $0$.
 
-* Fails with `STACK_UNDERFLOW` if the stack is empty.
+* Fails with `EMPTY_STACK` if the stack is empty.
```
SWvheerden, line 223: This change does not seem to be correct

SWvheerden (commented):
Make sure that the returned errors are all correct, they don't seem to be

stringhandler (commented):
Please add a note of the version of opcodes used and how the consensus code limits which opcodes are available at different block heights

src/RFC-0202_TariScriptOpcodes.md
hansieodendaal: Yip, that is my mistake, fixing all

SWvheerden (approved):
ACK

src/RFC-0202_TariScriptOpcodes.md
```diff
@@ -110,6 +109,17 @@ The full list of [Error codes](#error-codes) is given below.
 * The maximum length of a script's input is 1,024 bytes.
 * The maximum stack height is 255.
 
+### Opcode versions
+
+Base layer core consensus constants are linked to block height for each network, being it testnet, stagenet or mainnet, 
+and is backwards compatible, meaning a base node running updated consensus constants will also be able to validate the 
+blockchain for the previous version up to its last effective block height.
```
CjS77, line 116:
```suggestion
Base layer core consensus constants are linked to block height for each network, be it testnet, stagenet or mainnet, 
and are backwards compatible, meaning a base node running updated consensus constants will also be able to validate the 
blockchain for the previous version up to its last effective block height.
```

src/RFC-0202_TariScriptOpcodes.md
```diff
@@ -110,6 +109,17 @@ The full list of [Error codes](#error-codes) is given below.
...
+
+Opcode versioning is contained within the consensus constants and used to determine which opcodes should be effective 
+from which block height. As an example, `OpcodeVersion::V0` could be effective from the genysis block, 
+`OpcodeVersion::V1` may contain two additional opcodes and could be effective from height 1234, whereas 
+`OpcodeVersion::V2` may deprecate three other opcodes and be effective from height 21743.
```
CjS77, line 121:
```suggestion
Opcode versioning is contained within the consensus constants and is used to determine which opcodes are effective 
from which block height. As an example, `OpcodeVersion::V0` could be effective from the genesis block, 
`OpcodeVersion::V1` may contain two additional opcodes and could be effective from height 1234, whereas 
`OpcodeVersion::V2` may deprecate three other opcodes and be effective from height 21743.
```

CjS77 (approved):

PR #48, opened 2022-11-10
docs: update Emoji ID

SWvheerden:
Description
---
Move to stable
Update the EmojiID to encode the network address using the checksum
Do Grammarly check

### DIFF
--- a/src/RFC-0152_EmojiId.md
+++ b/src/RFC-0152_EmojiId.md
@@ -63,13 +63,13 @@ The most common practice for human beings to copy large numbers in cryptocurrenc
 encoding. The user will then typically scan (parts) of the string by eye to ensure that the value was transferred
 correctly.
 
-For Tari, we propose encoding values, the node ID in particular, for Tari, using emojis. The advantages of this approach are:
+For Tari, we propose encoding values, the node ID and network identifier in particular, for Tari, using emojis. The advantages of this approach are:
 
 * Emoji are more easily identifiable; and, if selected carefully, less prone to identification errors (e.g., mistaking an
   O for a 0).
 * The alphabet can be considerably larger than hexadecimal (16) or Base58 (58), resulting in shorter character sequences
   in the encoding.
-
+* Should be be able to detect if the address used belongs to the correct network. 
 ## The specification
 
 ### The emoji character map
@@ -81,8 +81,7 @@ list is the emoji map. For example,
 * ...
 * 🦊 => 255
 
-The emoji SHOULD be selected such that.
-
+The emoji SHOULD be selected such that:
 * Similar-looking emoji are excluded from the map. e.g. Neither 😁 or 😄 should be included. Similarly, the Irish and
   Côte d'Ivoire flags look very similar, and both should be excluded.
 * Modified emoji (skin tones, gender modifiers) are excluded. Only the "base" emoji is considered.
@@ -99,24 +98,29 @@ The emoji ID is a string of 33 symbols from the emoji alphabet. It uses this bit
 
 ```text
 +-----------------------------+--------------------------+
-|  Node public key (256 bits) | EncodedChecksum (8 bits) |
+|  Node public key (256 bits) | MaskedChecksum (8 bits)  |
 +-----------------------------+--------------------------+
 ```
 
  The emoji ID is calculated from a node public key serialized as 32 bytes (`B`) as follows:
 
 * Use the [DammSum](https://github.com/cypherstack/dammsum) algorithm with `k = 8` and `m = 32` to compute an 8-bit
-encoded checksum `eC` from `B`.
+masked checksum `'C` from `B` and network `N`.
 * Encode `B` into an emoji string.
-* Encode `eC` into an emoji string.
-* Concatenate `B` and `eC` as the emoji ID.
+* Encode `'C` into an emoji string.
+* Concatenate `B` and `'C` as the emoji ID.
+
+#### Checksum effectiveness
+It is important to note that by masking the checksum we technically reduce the effectiveness of the checksum. These emoji ids wont
+however be sent over the wire in raw form abd its far more likely that they wil just be copied and thus its important to verify the
+network of the wallet.
 
 ### Checksum 
 
-For the checksum we use the [DammSum](https://github.com/cypherstack/dammsum) algorithm with `k = 8` and `m = 32` to compute an 8-bit
-encoded checksum `C` from the public key `B`.
-To verify that the network that the public key comes from is the same one, we encode the checksum with the network. 
-We calculate the encoded checksum `eC` as `C XOR N` where `N` is the network defined as 8-bits.
+For the MaskedChecksum we use the [DammSum](https://github.com/cypherstack/dammsum) algorithm with `k = 8` and `m = 32` to compute an 8-bit
+Checksum `C` from the public key `B`.
+To assert that the network that the public key comes from is the expected network, we mask the checksum with the network. 
+We calculate the MaskedChecksum `'C` as `C XOR N` where `N` is the network defined as 8-bits.
 
 ### Decoding
 
@@ -124,8 +128,8 @@ One can extract the node public key from an emoji ID as follows:
 
 * Assert that the emoji ID contains exactly 33 valid emoji characters from the emoji alphabet. If not, return an error.
 * Decode the emoji ID as an emoji string by mapping each emoji character to a byte value using the emoji map, producing
-33 bytes. Let `B` be the first 32 bytes and `eC` be the last byte.
-* Calculate `C` by `C XOR N`.
+33 bytes. Let `B` be the first 32 bytes and `'C` be the last byte.
+* Calculate `C` by `'C XOR N`.
 * Use the DammSum validation algorithm on `B` to assert that `C` is a valid checksum. If not, return an error.
 * Attempt to deserialize `B` as a public key. If this fails, return an error. If it succeeds, return the public key.
 
@@ -133,8 +137,7 @@ One can extract the node public key from an emoji ID as follows:
 
 | Date         | Change                   | Author     |
 |:-------------|:-------------------------|:-----------|
-| 2022-11-10   | Finalise SHA-3 algorithm | SWvHeerden |
+| 2022-11-10   | Initial stable           | SWvHeerden |
 
 [Communication Node]: Glossary.md#communication-node
 [Node ID]: Glossary.md#node-id
-
--- a/src/RFC-0152_EmojiId.md
+++ b/src/RFC-0152_EmojiId.md
@@ -2,7 +2,7 @@
 
 ## Emoji Id specification
 
-![status: draft](theme/images/status-draft.svg)
+![status: stable](theme/images/status-stable.svg)
 
 **Maintainer(s)**:[Cayle Sharrock](https://github.com/CjS77)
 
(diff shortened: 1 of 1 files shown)


src/RFC-0152_EmojiId.md
```diff
@@ -83,46 +81,60 @@ list is the emoji map. For example,
 * ...
 * 🦊 => 255
 
-The emoji SHOULD be selected such that
+The emoji SHOULD be selected such that.
```
AaronFeickert, line 84: Probably want a colon `:` here instead.

src/RFC-0152_EmojiId.md
```diff
@@ -83,46 +81,60 @@ list is the emoji map. For example,
...
 
-The emoji ID is an emoji string of 33 symbols from the emoji alphabet. It uses this bitmap:
+The emoji ID is a string of 33 symbols from the emoji alphabet. It uses this bitmap:
 
 ```text
-+-----------------------------+-------------------+
-|  Node public key (256 bits) | Checksum (8 bits) |
-+-----------------------------+-------------------+
++-----------------------------+--------------------------+
+|  Node public key (256 bits) | EncodedChecksum (8 bits) |
```
AaronFeickert, line 102: I wouldn't use the term "encoded" for this, as I don't think it really conveys the nature of the modification. Maybe something like "masked" instead?

src/RFC-0152_EmojiId.md
```diff
@@ -83,46 +81,60 @@ list is the emoji map. For example,
...
++-----------------------------+--------------------------+
 ```
 
  The emoji ID is calculated from a node public key serialized as 32 bytes (`B`) as follows:
 
 * Use the [DammSum](https://github.com/cypherstack/dammsum) algorithm with `k = 8` and `m = 32` to compute an 8-bit
-checksum `C` from `B`.
+encoded checksum `eC` from `B`.
```
AaronFeickert, line 109: I'd be more specific about this, to avoid any confusion:
"... to compute an 8-bit masked checksum `C'` from `B` (described below)"

Otherwise, the casual reader might assume this is just the standard DammSum checksum.

src/RFC-0152_EmojiId.md
```diff
@@ -83,46 +81,60 @@ list is the emoji map. For example,
...
 * Encode `B` into an emoji string.
-* Encode `C` into an emoji string.
-* Concatenate `B` and `C` as the emoji ID.
+* Encode `eC` into an emoji string.
+* Concatenate `B` and `eC` as the emoji ID.
 
-#### Decoding
+### Checksum 
+
+For the checksum we use the [DammSum](https://github.com/cypherstack/dammsum) algorithm with `k = 8` and `m = 32` to compute an 8-bit
```
AaronFeickert, line 116: "For the masked checksum we use..."

src/RFC-0152_EmojiId.md
```diff
@@ -83,46 +81,60 @@ list is the emoji map. For example,
...
+encoded checksum `C` from the public key `B`.
+To verify that the network that the public key comes from is the same one, we encode the checksum with the network.
```
AaronFeickert, line 118: I would use a term like "assert" instead of "verify", since verification seems to imply part of the decoding and validation process.

src/RFC-0152_EmojiId.md
```diff
@@ -83,46 +81,60 @@ list is the emoji map. For example,
...
 ```text
-+-----------------------------+-------------------+
-|  Node public key (256 bits) | Checksum (8 bits) |
-+-----------------------------+-------------------+
++-----------------------------+--------------------------+
+|  Node public key (256 bits) | EncodedChecksum (8 bits) |
++-----------------------------+--------------------------+
 ```
 
  The emoji ID is calculated from a node public key serialized as 32 bytes (`B`) as follows:
```
AaronFeickert, line 106: Should mention that the 8-bit network encoding is part of the algorithm input here, along with the public key. Otherwise it's not clear where it comes into play.

src/RFC-0152_EmojiId.md
```diff
@@ -83,46 +81,60 @@ list is the emoji map. For example,
...
+### Checksum 
+
+For the checksum we use the [DammSum](https://github.com/cypherstack/dammsum) algorithm with `k = 8` and `m = 32` to compute an 8-bit
+encoded checksum `C` from the public key `B`.
+To verify that the network that the public key comes from is the same one, we encode the checksum with the network. 
+We calculate the encoded checksum `eC` as `C XOR N` where `N` is the network defined as 8-bits.
+
+### Decoding
 
 One can extract the node public key from an emoji ID as follows:
```
AaronFeickert, line 123: "... from an emoji ID and expected network `N` as follows:"

src/RFC-0152_EmojiId.md
AaronFeickert: I would also avoid the use of multiple symbols for the variable name `eC`, but that's just a nit and a personal preference.

AaronFeickert (changes requested):

AaronFeickert:
Should have also mentioned in the review that this approach technically reduces the effectiveness of the emoji ID. If a different network identifier is used during decoding and a transmission error occurs that mangles the emoji ID in the right way, it's possible for the checksum to pass when it shouldn't. This should probably be noted somewhere.

AaronFeickert:
It might also be useful to indicate in the narrative _why_ the network byte is used, and to highlight that addresses intended for different use cases should have different network bytes.

Further, if it's in scope, it might be helpful to specify which network bytes Tari will use, or at least which network differentiations will be needed (project, address type, network).

src/RFC-0152_EmojiId.md
```diff
@@ -83,46 +81,63 @@ list is the emoji map. For example,
...
+masked checksum `'C` from `B` and network `N`.
 * Encode `B` into an emoji string.
-* Encode `C` into an emoji string.
-* Concatenate `B` and `C` as the emoji ID.
+* Encode `'C` into an emoji string.
+* Concatenate `B` and `'C` as the emoji ID.
+
+#### Checksum effectiveness
+It is important to note that by masking the checksum we technically reduce the effectiveness of the checksum. These emoji ids wont
+however be sent over the wire in raw form abd its far more likely that they wil just be copied and thus its important to verify the
```
stringhandler, line 115:
```suggestion
however, be sent over the wire in raw form and its far more likely that they will just be copied thus it's important to verify the
```

stringhandler (approved):

stringhandler:
Just waiting for approval from @AaronFeickert, then can merge

src/RFC-0152_EmojiId.md
```diff
@@ -48,8 +48,7 @@ technological merits of the potential system outlined herein.
 
 ## Goals
 
-This document describes the specification for Emoji Ids. Emoji Ids are encoded node ids used for humans to easily verify
-peer node addresses.
+This document describes the specification for Emoji Ids. Emoji Ids are encoded node ids used for humans to verify peer node addresses easily.
```
CjS77, line 51:
```suggestion
This document describes the specification for Emoji Ids. Emoji Ids are encoded node ids used for humans to verify peer node addresses easily and for machines to verify that the address is being used in the correct context.
```

src/RFC-0152_EmojiId.md
```diff
@@ -60,21 +59,20 @@ None
...
 public key. Both the node id and public key are simple large integer numbers.
 
-The most common practice for human beings to copy large numbers in cryptocurrency software is to scan a QR code or copy
-and paste a value from one application to another. These numbers are typically encoded using hexadecimal or Base58
+The most common practice for human beings to copy large numbers in cryptocurrency software is scanning a QR code or copying and pasting a value from one application to another. These numbers are typically encoded using hexadecimal or Base58
 encoding. The user will then typically scan (parts) of the string by eye to ensure that the value was transferred
 correctly.
 
-For Tari, we propose encoding values, the node ID in particular, using emoji. The advantages of this approach are:
+For Tari, we propose encoding values, the node ID and network identifier in particular, for Tari, using emojis. The advantages of this approach are:
```
CjS77, line 66: The network id is masking the checksum, so is not encoded.

src/RFC-0152_EmojiId.md
```diff
@@ -83,46 +81,63 @@ list is the emoji map. For example,
 * ...
 * 🦊 => 255
 
-The emoji SHOULD be selected such that
-
-* Similar looking emoji are excluded from the map. e.g. Neither 😁 or 😄 should be included. Similarly the Irish and
-  Côte d'Ivoirean flags look very similar, and both should be excluded.
+The emoji SHOULD be selected such that:
```
CjS77, line 84: Change language to past continuous tense.

```suggestion
The emoji have been selected such that:
```

src/RFC-0152_EmojiId.md
```diff
@@ -83,46 +81,63 @@ list is the emoji map. For example,
...
+* Similar-looking emoji are excluded from the map. e.g. Neither 😁 or 😄 should be included. Similarly, the Irish and
+  Côte d'Ivoire flags look very similar, and both should be excluded.
```
CjS77, line 86: May as add the table of emoji here and their encoding values.

src/RFC-0152_EmojiId.md
```diff
@@ -83,46 +81,63 @@ list is the emoji map. For example,
...
 * Encode `B` into an emoji string.
-* Encode `C` into an emoji string.
-* Concatenate `B` and `C` as the emoji ID.
+* Encode `'C` into an emoji string.
+* Concatenate `B` and `'C` as the emoji ID.
+
+#### Checksum effectiveness
+It is important to note that by masking the checksum we technically reduce the effectiveness of the checksum. These emoji ids wont
+however, be sent over the wire in raw form and its far more likely that they will just be copied thus it's important to verify the
+network of the wallet.
```
CjS77, line 116: Specifically, if you want to get technical, there's a ~4/256 = 1/64 chance that a transcription mistake leads to the incorrect network/version combination matching the checksum (assuming 4 ver/net combos).

CjS77 (commented):
A few suggestions

AaronFeickert:
I still think the encoding and algorithm descriptions aren't strictly correct as written (but maybe this is being too picky).

In the encoding algorithm, the step
```
Use the DammSum algorithm with `k = 8` and `m = 32` to compute an 8-bit masked checksum `'C` from `B` and network `N`.
```
isn't quite right, because DammSum doesn't produce masked checksums. Further, the algorithm doesn't specify that `N` is an input to it.

In the decoding algorithm, it also isn't specified that `N` is a required input.

The missing step in the encoding algorithm is to get the unmasked checksum from DammSum and then apply the masking. I'm a fan of making algorithm descriptions as clear as possible; if I were a first-time reader, I'd probably need to check an implementation to confirm what's actually going on (which reduces the effectiveness of the RFC).

To be specific, I'd change the encoding algorithm steps to be something like this:
```
The emoji ID is produced from a node public key `B` (serialized as 32 bytes) and a network identifier `N` (serialized as 8 bits) as follows:
- Use the DammSum algorithm with `k = 8` and `m = 32` to compute an 8-bit checksum `C` using `B` as input.
- Compute the masked checksum `C' = C XOR N`.
- Encode `B` into an emoji string using the emoji map.
- Encode `C'` into an emoji character using the emoji map.
- Concatenate `B` and `C'` as the emoji ID.
```

And I'd change the "preamble" of the decoding algorithm to be something like this:
```
The node public key is obtained from an emoji ID and a network identifier `N` (serialized a 8 bits) as follows:
```

PR #49, opened 2022-11-10
docs: rfc-0313 describe validator node registration process

sdbondi:
Description
---

Motivation and Context
---

How Has This Been Tested?
---


src/RFC-0313_VNRegistration.md
```diff
@@ -0,0 +1,344 @@
...
+| Name                           | Value              | Description                                                                                |
+|:-------------------------------|:-------------------|:-------------------------------------------------------------------------------------------|
+| `EpochLength`                  | 60 Blocks (~2h)    | The number of blocks in an epoch                                                           |
+| `VNRegistrationValidityPeriod` | 20 Epochs (~40hrs) | The number of epochs that a validator node registration is valid                           |
+| `VNRegDepositAmount`           | TBD Tari           | The minimum amount that must be spent to create a valid `ValidatorNodeRegistration` [UTXO] |
+| `VNRegLockHeight`              | 10 Epochs          | The lock height that must be set on every `ValidatorNodeRegistration` [UTXO]               |
+| `VNShardShuffleInterval`       | 100 Epochs         | The interval that a validator node shard key is shuffled                                   |
+| `VNRegInstateDelay`            | 2 Epochs           | The number of epochs to delay for before instating a new VN                                |
```
stringhandler, line 95: Suggestion: Activation instead of instate
```suggestion
| `VNRegActivationDelay`            | 2 Epochs           | The number of epochs to delay for before activating a new VN                                |
```

src/RFC-0313_VNRegistration.md
```diff
@@ -0,0 +1,344 @@
...
+OP_END_IF
+```
+
+By submitting this [UTXO] the validator node operator is committing to providing a highly-available node for `VNRegistrationValidityPeriod` epochs.
+
+A validator node operator MAY re-register their `VN_Public_Key` before the `VNRegistrationValidityPeriod` epoch is reached, OPTIONALLY 
+spending the previous `ValidatorNodeRegistration` [UTXO]. If the previous `ValidatorNodeRegistration` [UTXO] has not expired and
+a new `ValidatorNodeRegistration` [UTXO] is submitted, the new `ValidatorNodeRegistration` [UTXO] supersedes the previous one.
+
+This allows a validator to re-register earlier than the lock-height as long as they are willing to commit
```
stringhandler, line 132: Is this possibly not the end of the sentence?

src/RFC-0313_VNRegistration.md
```diff
@@ -0,0 +1,344 @@
...
+- the validator MUST reject instructions for epoch 12.
+
+Point (b) - the start of epoch 12: 
+- the validator nodeset is final for epoch 12.
+- the active epoch remains at 11.
+- the validator MUST accept instructions from epoch 11 and 12.
+
+Point (c) - the transition point for epoch 12:
+- the active epoch is now 12
+- the validator MUST accept instructions from 11.
```
stringhandler, line 215:
```suggestion
- the validator MUST accept instructions from epoch 11 and 12.
```

stringhandler: it is not clear what should happen between C and D. Is it the rules from C or the rules from D?

src/RFC-0313_VNRegistration.md
```diff
@@ -0,0 +1,344 @@
...
+- it is assumed at this point that almost all ($3f + 1$) nodes are now on epoch 12
+
+Point (d) - the epoch mid-point:
+- the active epoch is 12.
+- the validator MUST reject instructions from epoch 11.
+- the validator MUST accept instructions from epoch 12.
+
+#### Validator Node Set Definition
+
+The function $\text{get_vn_set}(\epsilon_n) \rightarrow \vec{S}$ that returns an ordered vector $\vec{S}$ of `VN_Shard_Key`s that are registered for the epoch $\epsilon_n$.
```
stringhandler, line 225: Not sure if this will show up in the mdbook version, but Github shows an error rendering this:
image

stringhandler (commented):
I think there needs to be more detail on the shuffling of  nodes per epoch to ensure that only X% of nodes are shuffled

src/RFC-0313_VNRegistration.md
sdbondi: oh that's annoying, yeah it renders fine in mdbook

src/RFC-0313_VNRegistration.md
sdbondi: From point (c) after the transition point, only epoch 12 is allowed - so we allow a few blocks after the epoch where nodes can be behind. I should have removed point (d) actually, there is no change. The mid point isnt really used - you can set the transition point to be half of the epoch which is the same thing.

src/RFC-0313_VNRegistration.md
```diff
@@ -0,0 +1,343 @@
...
+and processed by all [base node]s. Moreover, the validators may poll the base layer only every few seconds, further increasing the latency for 
+validators to become aware of the state. This means a single strict validator epoch change-over point will almost always cause liveness failures at 
+and after the epoch transition.
+
+To address this, we define a `VNEpochGracePeriod` where both the previous and current epoch are accepted. This value, in blocks, must allow enough
+time for all validators to become aware of the [base layer] state for the epoch.
+
+To illustrate, consider the following view of a [base layer] chain. We mark 3 views of the chain.
+
+```text
```
SWvheerden, line 186: We need to take a look here wat happens with leader changes, on point b, we need to make a rule surrounding what leader to listen to, we might have two leaders in E11 and E12, both honest. 

which 2 do the nodes listen to? Their local epoch?

SWvheerden (commented):
looks good so far
We need to address why its important to add why we want to limit VNC changes to a set %

src/RFC-0313_VNRegistration.md
```diff
@@ -0,0 +1,343 @@
...
+## Validator Registration
+
+A validator node operator wishing to participate in [DAN] consensus MUST generate a [Ristretto] keypair `<VN_Public_Key, VN_Secret_Key>` that serves as a stable [DAN] identity and a signing key for L2 consensus messages.
+The `VN_Public_Key` MUST be registered on the base layer by submitting a `ValidatorNodeRegistration` [UTXO] which allows the validator to participate in [DAN] consensus for `VNRegistrationValidityPeriod`.
+
+The published `ValidatorNodeRegistration` [UTXO] has these requirements:
+1. MUST contain a valid [Schnorr signature] that proves knowledge of the `VN_Secret_Key` 
+    * The signature challenge is defined as $e = H(R \mathbin\Vert P \mathbin\Vert m)$
+    * $R$ is a public nonce, $P$ is the public `VN_Public_Key` 
+    * $m$ is TODO (previous shard key?, nothing?)
```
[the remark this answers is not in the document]
stringhandler: I suggest using the commitment of the UTXO.

src/RFC-0313_VNRegistration.md
```diff
@@ -0,0 +1,343 @@
...
+```
+
+Point (a)
+- the validator node set is incomplete for epoch 12.
+- the active epoch is 11.
+- the validator MUST reject instructions for epoch 12.
+
+Point (b) - the start of epoch 12: 
+- the validator nodeset is final for epoch 12.
+- the active epoch remains at 11.
```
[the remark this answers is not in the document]
stringhandler: Maybe: VN leaders SHOULD propose using epoch 12, but MAY choose to propose using epoch 11

src/RFC-0313_VNRegistration.md
stringhandler: As long as they only vote for one message out of the two leaders, it is fine

src/RFC-0313_VNRegistration.md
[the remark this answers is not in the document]
sdbondi: If nodes are behind they will reject 12, 12 kicks in after the transition period.

sdbondi:
We may have to change how this works to achieve a precise shuffle rate. Currently, we can choose a high `VNShardShuffleInterval` to get probabilistically close to 5% on average. 

Another approach may be to randomly but deterministically sample 5% of the current validator set using a seeded RNG, with some known entropy (e.g. prev_block_hash or utxo_mr) as the seed. Once selected, the shard key is shuffled $S_i = H(V_i || \eta)$ where $\eta$ (eta) is the same shared entropy.

src/RFC-0313_VNRegistration.md
```diff
@@ -319,6 +298,12 @@ fn derive_shard_key(prev_shard_key: Option<ShardId>, vn_key: U256, epoch: u64, i
 }
 ```
 
+Only a random fraction of validators will be re-assigned shard keys per epoch and that fraction will not be shuffled again
+for `VNShardShuffleInterval` epochs. Although the exact number of validators that shuffle per epoch varies, on average 
+the `VNShardShuffleInterval` should aim to shuffle around 5% of the network at every epoch. This is to ensure that the
+number of shuffling validators is far from $\frac{1}{3}$ of the validator set, as this could break liveness and safety
```
stringhandler, line 304:
```suggestion
number of shuffling validators is much less than $\frac{1}{3}$ of the validator set, as this could break liveness and safety 
```

stringhandler (dismissed):

CjS77 (approved):

PR #65, opened 2022-11-14
fix: rfc-0121 is deprecated

agubarev:
Description
---
Moved old RFC-0121 describing Consensus Encoding to the deprecated section.

Motivation and Context
---
Initial implementation is deprecated in favor of `borsh`

How Has This Been Tested?
---

### DIFF
--- a/src/RFC-0201_TariScript.md
+++ b/src/RFC-0201_TariScript.md
@@ -85,9 +85,8 @@ Some smart contract features are possible, or partly possible in vanilla [Mimble
 - Atomic swaps 
 - Hash time-locked contracts
 
-This RFC makes the case that if Tari were to implement a scripting language similar to Bitcoin script, then all of these
-use cases will collapse and can be achieved under a single set of (relatively minor) modifications and additions to the
-current Tari and Mimblewimble protocol.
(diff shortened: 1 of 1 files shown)


src/RFC-0120_Consensus.md
```diff
@@ -360,7 +360,7 @@ The block MUST be transmitted in canonical ordering. The advantage of this appro
 done by the whole network, and verification of sorting is exceptionally cheap.
 
 Transaction outputs are sorted lexicographically by the byte representation of their Pedersen commitment i.e. `\\(k \cdot G + v \cdot H\\)`.
-Transaction inputs are sorted lexicographically by the [hash of the output](RFC-0121_ConsensusEncoding.md#transaction-output) that is spent by the input.
+Transaction inputs are sorted lexicographically by the [hash of the output](RFCD-0121_ConsensusEncoding.md#transaction-output) that is spent by the input.
```
CjS77, line 363: This should be updated to what the code is doing. Might be unchanged, but then remove the link to a deprecated RFC.

agubarev: I've removed 2 links pointing to this deprecated document, here and from the RFC-0131/Mining.
Will have to update https://rfc.tari.com/RFC-0131_Mining.html once new implementation starts working.

PR #67, opened 2022-11-15
feat: update rfc-0001 and stabilise

CjS77:
The aim of this proposal is to provide a very high-level perspective of the moving parts of the Tari protocol.

it has been updated to reflect changes in the DAN  consensus algorithm; i.e. Cerberus


src/RFC-0001_overview.md
```diff
@@ -102,134 +102,112 @@ We can't have fast, cheap digital assets and also highly secure and decentralize
...
-1. It manages and enforces the accounting and consensus rules of the base Tari (XTR) token. This includes standard payments,
-   and simple smart contracts such as cross-chain atomic swaps.
-2. It maintains the Digital Asset register. This includes functionality for side-chain management, checkpoints, and transfer
-   of funds into and out of side-chains.
-4. Maintain a register of smart contract templates. This allows users to verify that digital assets are running the code that they 
-   expect and includes functionality like version tracking.
+### Interplay of the layers
 
-In general, the base layer knows _nothing about the specifics of what is happening on the side-chains_. It only cares
+In general, the base layer knows _nothing about the specifics of what is happening on the side-chain_. It only cares
```
stringhandler, line 177: It does know about Epochs, Validator nodes and templates, but I understand what you're saying

stringhandler (approved):

src/RFC-0001_overview.md
CjS77: And even so, The consensus code only has the vaguest sense. It's metadata attached to std Tari txs. How much validation are we planning on doing on that metadata in the consensus validation code?

PR #81, opened 2022-11-25
fix: rfc-0121 is deprecated

agubarev:
Description
---
Moved old RFC-0121 describing Consensus Encoding to the deprecated section.

Motivation and Context
---
Initial implementation is deprecated in favor of `borsh`

How Has This Been Tested?
---

### DIFF
--- a/src/RFC-0131_Mining.md
+++ b/src/RFC-0131_Mining.md
@@ -119,8 +119,8 @@ Tari's proof-of-work mining algorithm is summarized below:
 
 ### Tari mining hash
 
-First, the block header is hashed with the 256-bit Blake2b hashing algorithm based using the approach described in
-[consensus encoding](RFC-0121_ConsensusEncoding). The fields are hashed in the order:
+First, the block header is hashed with the 256-bit Blake2b hashing algorithm using the approach described in
+consensus encoding. The fields are hashed in the order:
 - version
 - previous header hash
 - timestamp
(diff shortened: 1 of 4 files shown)


src/RFC-0131_Mining.md
```diff
@@ -119,8 +119,8 @@ Tari's proof-of-work mining algorithm is summarized below:
 
 ### Tari mining hash
 
-First, the block header is hashed with the 256-bit Blake2b hashing algorithm based using the approach described in
-[consensus encoding](RFC-0121_ConsensusEncoding). The fields are hashed in the order:
+First, the block header is hashed with the 256-bit Blake2b hashing algorithm using the approach described in
+consensus encoding. The fields are hashed in the order:
```
CjS77, line 123: Here's the relevant code snippet:
```rust
    pub fn mining_hash(&self) -> FixedHash {
        DomainSeparatedConsensusHasher::<BlocksHashDomain>::new("block_header")
            .chain(&self.version)
            .chain(&self.height)
            .chain(&self.prev_hash)
            .chain(&self.timestamp)
            .chain(&self.input_mr)
            .chain(&self.output_mr)
            .chain(&self.output_mmr_size)
            .chain(&self.witness_mr)
            .chain(&self.kernel_mr)
            .chain(&self.kernel_mmr_size)
            .chain(&self.total_kernel_offset)
            .chain(&self.total_script_offset)
            .finalize()
            .into()
    }
    ```

It looks like we've missed block height. Also, we're using domain-separated hashing. I don't think there's an RFC on this, but you could link to the tari-crypto crate where it's implemented. The domain we're using is `"com.tari.base_layer.core.blocks"`

CjS77 (commented):
Picked up one omission. Otherwise GTG

src/RFC-0131_Mining.md
CjS77: Fixed in different commit

PR #95, opened 2023-04-14
feat: updated CAPK signature and security proofs

AaronFeickert:
Description
---
Moves the description of commitment and public key (CAPK) signatures from RFC-0201 to a new RFC-0182. Adds a variation for commitments with revealed value. Includes new security proofs for both constructions.

Motivation and Context
---
There has been discussion about ways to make transactions more efficient in the case where the sender does not wish to hide a commitment's value. It turns out that CAPK signatures can be modified such a verifier can assert a commitment binds to a revealed value, which removes the need for range proof verification.

This PR does a few things. First, it moves the existing CAPK signature description from RFC-0201 to its own new RFC-0182. Then, it splits the description into two protocols: one that hides the commitment value, and another that reveals it. Finally, it adds a formal proof of security for each of the protocols; each proof shows the corresponding protocol is correct, special sound, and special honest-verifier zero knowledge.

How Has This Been Tested?
---
The changes build and appear to render correctly. Updated and new links work as expected.

### DIFF
--- a/src/RFC-0182_CommitmentSignatures.md
+++ b/src/RFC-0182_CommitmentSignatures.md
@@ -1,0 +1,183 @@
+# RFC-0182/CommitmentSignatures
+
+## Commitment and public key signatures
+
+![status: draft](theme/images/status-draft.svg)
+
+**Maintainer(s)**: [Aaron Feickert](https://github.com/AaronFeickert)
+
+# Licence
+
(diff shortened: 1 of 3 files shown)


CjS77 (approved):
LGTM

CjS77:
ACK

hansieodendaal:
Awesome

PR #98, opened 2023-05-17
feat: hardware wallet transactions

SWvheerden:
Description
---
Intial draft for hardware wallet transactions. 

Motivation and Context
---

How Has This Been Tested?
---

### DIFF
--- a/src/RFC-0205_HardwareTransactions.md
+++ b/src/RFC-0205_HardwareTransactions.md
@@ -68,7 +68,7 @@ get around these limitations to have fully functional secure hardware wallet int
 
 ## Background
 
-Vanilla [Mimblewimble] only has a single secret per [UTXO], the blinding factor (\\( k_i \\) ), the Tari protocol extends the [Mimblewimble] protocol to include scripting in the form of [TariScript]. This adds a second secret per [UTXO] called the script_key (\\( k_s \\) ). In practical terms this means that while vanilla [Mimblewimble] only requires that a wallet wishing to spend an [UTXO], prove knowledge of (\\( k_i \\) ) by producing the kernel signature, this is not sufficient for Tari. A Tari wallet must also prove knowledge of the script key (\\( k_s \\) ), by producing the script signature. 
+Vanilla [Mimblewimble] only has a single secret per [UTXO], the blinding factor (\\( k_i \\) ). The Tari protocol extends the [Mimblewimble] protocol to include scripting in the form of [TariScript]. This adds a second secret per [UTXO] called the script_key (\\( k_s \\) ). In practical terms this means that while vanilla [Mimblewimble] only requires that a wallet wishing to spend an [UTXO], prove knowledge of (\\( k_i \\) ) by producing the kernel signature, this is not sufficient for Tari. A Tari wallet must also prove knowledge of the script key (\\( k_s \\) ), by producing the script signature. 
 
 ## Requirements
 
@@ -117,10 +117,10 @@ This can be done by the helper asking the signer for a public key. And advertisi
 The helper can scan the blockchain for this public key. 
 
 ### Receiving one-sided-stealth
-`TODO`
+Not yet possible with this this design.
 
 ### Output recovery
-When creating outputs the wallet encrypts the blinding factor \\(k_i \\) and value \\( v \\) with \\( k_H \\).
+When creating outputs the wallet encrypts the blinding factor \\(k_i \\) and value \\( v \\) with \\( k_H \\). This is encrypted using extended-nonce AEAD using a random nonce and authenticated decryption.
 Because the key \\( k_H \\)  is calculated from the seed phrase of the signer, this will be the same each time. The helper can try to decrypt each scanned output, when it is successful it knows it has found its own output. 
 The helper can validate that the commitment is correct using the blinding factor \\(k_i \\) and value \\( v \\). It can also validate (\\( K_S)\\) corresponds to (\\( k_i, A \\) )
 
--- a/src/RFC-0205_HardwareTransactions.md
+++ b/src/RFC-0205_HardwareTransactions.md
@@ -77,6 +77,7 @@ To properly implement hardware wallets we need the following requirements to be 
 * Users need to verify transaction properties when signing for transactions.
 * The user must be able to receive transactions without having to authorize them on the hardware wallet.
 * The user must be able to receive transactions without having the Hardware wallet attached.
+* The hardware device implementation should comply with the best practices and/or security recommendations of the provider.
 
 ## Implementation
 
--- a/src/RFC-0205_HardwareTransactions.md
+++ b/src/RFC-0205_HardwareTransactions.md
@@ -87,7 +87,7 @@ Normal transactions have only a single entity, the wallet which controls all sec
 ### Process Overview
 By splitting the ownership of the [UTXO]'s secrets by assigning knowledge of only the script key (\\( k_s \\) ) to the signer, we can lift much of the heavy cryptography like bulletproof creation to the helper device by exposing (\\( k_i \\) ) to it. By looking at how [one-sided-stealth] transactions are created, we can construct the script key in such a way that the helper can calculate the public script key, but cannot calculate the private script key.
 
-All hardware wallet created [UTXO]s created will be created with a script `PushPubkey(K_S)`. The key (\\( k_s \\) ) is created as follows:
+All hardware wallet created [UTXO]s will contain a script `PushPubkey(K_S)`. The key (\\( k_s \\) ) is created as follows:
 $$
 \begin{align}
 k_S &= H(k_i) + a \\\\
--- a/src/RFC-0205_HardwareTransactions.md
+++ b/src/RFC-0205_HardwareTransactions.md
@@ -81,18 +81,18 @@ To properly implement hardware wallets we need the following requirements to be 
 ## Implementation
 
 ### Entities
-Normal transactions have only a single entity, the wallet which controls all secrets and transactions. But with hardware wallets, we need to define two distinct types:
+Normal transactions have only a single entity, the wallet which controls all secrets and transactions. But with hardware wallets, we need to define two distinct entities:
 * Signer: This is the entity that keeps the secrets for the transactions and approves them, aka the hardware wallet.
 * Helper: This entity is the program that helps the signer construct the transaction, send it over the network and scan the network, aka wallet.
 
 ### Process Overview
-By splitting the ownership of the [UTXO]'s secret by assigning knowledge of only the script key (\\( k_s \\) ) to the signer, we can lift much of the heavy cryptography like bulletproof creation to the helper device by exposing (\\( k_i \\) ) to it. By looking at how [one-sided-stealth] transactions are created, we can construct the script key in such a way that the helper can calculate the public script key, but cannot calculate the private script key.
+By splitting the ownership of the [UTXO]'s secrets by assigning knowledge of only the script key (\\( k_s \\) ) to the signer, we can lift much of the heavy cryptography like bulletproof creation to the helper device by exposing (\\( k_i \\) ) to it. By looking at how [one-sided-stealth] transactions are created, we can construct the script key in such a way that the helper can calculate the public script key, but cannot calculate the private script key.
 
 All hardware wallet created [UTXO]s created will be created with a script `PushPubkey(K_S)`. The key (\\( k_s \\) ) is created as follows:
 $$
 \begin{align}
-k_S &= k_i + a \\\\
-K_S &= k_i \cdot G + A
+k_S &= H(k_i) + a \\\\
+K_S &= H(k_i) \cdot G + A
 \end{align}
 $$
 
--- a/src/RFC-0205_HardwareTransactions.md
+++ b/src/RFC-0205_HardwareTransactions.md
@@ -7,7 +7,6 @@
 **Maintainer(s)**: [SW van Heerden](https://github.com/swvheerden)
 
 
-\newcommand{\hash}[1]{\mathrm{H}\bigl({#1}\bigr)}
 
 # Licence
 
@@ -105,7 +104,7 @@ Helper asks signer for master helper key (\\( k_H \\) ).
 This key is derived from the signer seed phrase.
 
 ### Transaction receiving
-When a transaction is received the helper constructs the new [UTXO] with its Rangeproof. Choosing a new (\\( k_i \\) ) for the [UTXO], it calculates a new \\( K_S \\). It attaches the script `PushPubkey(K_S)` to output.
+When a transaction is received the helper constructs the new [UTXO] with its Rangeproof. Choosing a new ( \\( k_i \\) ) for the [UTXO], it calculates a new \\( K_S \\). It attaches the script `PushPubkey(K_S)` to output.
 
 ### Transaction sending
 When the user wants to send a transaction, the helper retrieves the desired [UTXO]. The helper asks the signer to sign the transaction. 
@@ -121,9 +120,9 @@ The helper can scan the blockchain for this public key.
 `TODO`
 
 ### Output recovery
-When creating outputs the wallet encrypts the blinding factor (\\k_i \\) and value \\( v \\) with (\\( k_H \\) ).
-Because the key (\\( k_H \\) ) is calculated from the seed phrase of the signer, this will be the same each time. The helper can try to decrypt each scanned output, when it is successful it knows it has found its own output. 
-The helper can validate that the commitment is correct using the blinding factor (\\k_i \\) and value \\( v \\). It can also validate (\\( K_S)\\) ) corresponds to (\\( k_i, A \\) )
+When creating outputs the wallet encrypts the blinding factor \\(k_i \\) and value \\( v \\) with \\( k_H \\).
+Because the key \\( k_H \\)  is calculated from the seed phrase of the signer, this will be the same each time. The helper can try to decrypt each scanned output, when it is successful it knows it has found its own output. 
+The helper can validate that the commitment is correct using the blinding factor \\(k_i \\) and value \\( v \\). It can also validate (\\( K_S)\\) corresponds to (\\( k_i, A \\) )
 
 ## Security
 Because the script key is required for spending, it is the only key that needs to be kept secret. 
--- a/src/RFC-0205_HardwareTransactions.md
+++ b/src/RFC-0205_HardwareTransactions.md
@@ -1,6 +1,6 @@
-# RFC-0205/ledger transactions
+# RFC-0205/Hardware transactions
 
-## Ledger addresses
+## Hardware transactions
 
 ![status: stable](theme/images/status-stable.svg)
 
--- a/src/RFC-0205_HardwareTransactions.md
+++ b/src/RFC-0205_HardwareTransactions.md
@@ -76,6 +76,7 @@ To properly implement hardware wallets we need the following requirements to be 
 * No UTXO can be spent without a user physically approving the transaction on the hardware wallet.
 * Users need to verify transaction properties when signing for transactions.
 * The user must be able to receive transactions without having to authorize them on the hardware wallet.
+* The user must be able to receive transactions without having the Hardware wallet attached.
 
 ## Implementation
 
--- a/src/RFC-0205_LedgerTransactions.md
+++ b/src/RFC-0205_LedgerTransactions.md
@@ -69,7 +69,7 @@ get around these limitations to have fully functional secure hardware wallet int
 
 ## Background
 
-Vanilla [Mimblewimble] only has a single secret per [UTXO], the blinding factor (\\( k_i \\) ), the Tari protocol extends the [Mimblewimble] protocol to include scripting in the form of [TariScript]. This adds a second secret per [UTXO] called the script_key (\\( k_s \\) ). In practical terms this means that while vanilla [Mimblewimble] only requires that a wallet wishing to spend an [UTXO], prove knowledge of (\\( k_i \\) ) by singing the kernel signature, this is not sufficient for Tari. A Tari wallet must also prove knowledge of the script key (\\( k_s \\) ), by signing the script signature. 
+Vanilla [Mimblewimble] only has a single secret per [UTXO], the blinding factor (\\( k_i \\) ), the Tari protocol extends the [Mimblewimble] protocol to include scripting in the form of [TariScript]. This adds a second secret per [UTXO] called the script_key (\\( k_s \\) ). In practical terms this means that while vanilla [Mimblewimble] only requires that a wallet wishing to spend an [UTXO], prove knowledge of (\\( k_i \\) ) by producing the kernel signature, this is not sufficient for Tari. A Tari wallet must also prove knowledge of the script key (\\( k_s \\) ), by producing the script signature. 
 
 ## Requirements
 
@@ -82,8 +82,8 @@ To properly implement hardware wallets we need the following requirements to be 
 
 ### Entities
 Normal transactions have only a single entity, the wallet which controls all secrets and transactions. But with hardware wallets, we need to define two distinct types:
-* Signer: This is the device that keeps the secrets for the transactions and approves them, aka the hardware wallet.
-* Helper: This device is the program that helps the signer construct the transaction and send it over the network, aka wallet.
+* Signer: This is the entity that keeps the secrets for the transactions and approves them, aka the hardware wallet.
+* Helper: This entity is the program that helps the signer construct the transaction, send it over the network and scan the network, aka wallet.
 
 ### Process Overview
 By splitting the ownership of the [UTXO]'s secret by assigning knowledge of only the script key (\\( k_s \\) ) to the signer, we can lift much of the heavy cryptography like bulletproof creation to the helper device by exposing (\\( k_i \\) ) to it. By looking at how [one-sided-stealth] transactions are created, we can construct the script key in such a way that the helper can calculate the public script key, but cannot calculate the private script key.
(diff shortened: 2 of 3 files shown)


src/RFC-0205_LedgerTransactions.md
```diff
@@ -0,0 +1,148 @@
...
+The helper can attach the correct signatures to the [UTXO]s and ship the transaction.
+
+### Receiving normal one-sided transaction
+This can be done by the helper asking the signer for a public key. And advertising this public key as the destination public key for a 1-sided transaction. 
+The helper can scan the blockchain for this public key. 
+
+### Receiving one-sided-stealth
+`TODO`
+
+### Output recovery
```
AaronFeickert, line 124: It may be useful to describe how the helper can check the validity of an output's commitment and script public key after decrypting `v` and `k_i`:
- The helper checks that `v` and `k_i` are a valid opening of the commitment by reconstructing it
- The helper then checks that `K_s` corresponds to `k_i` and `A`

This process can detect malformed outputs that would not be spendable.

AaronFeickert (changes requested):

src/RFC-0205_LedgerTransactions.md
```diff
@@ -0,0 +1,148 @@
...
+Because the key (\\( k_H \\) ) is calculated from the seed phrase of the signer, this will be the same each time. The helper can try to decrypt each scanned output, when it is successful it knows it has found its own output. 
+
+## Security
+Because the script key is required for spending, it is the only key that needs to be kept secret. 
+The following table explains what an attacker can do upon learning a key from the helper
+
+| key         | Worst case scenario          | 
+|:------------|:-----------------------------|
+| \\( k_H \\) | Can view all transactions and values made by the wallet |
+| \\( k_i \\) | Can try and brute force the value of the transaction |
```
AaronFeickert, line 135: This can also occur if `K_i` were leaked.

AaronFeickert:
What should be the relationship between this RFC and the existing [RFC-0204](https://rfc.tari.com/RFC-0204_TariScriptExamples.html) (which discusses interactive and "legacy" one-sided outputs) and [RFC-0203](https://rfc.tari.com/RFC-0203_StealthAddresses.html) (which discusses stealth one-sided outputs)?

src/RFC-0205_LedgerTransactions.md
```diff
@@ -0,0 +1,148 @@
...
+## Background
+
+Vanilla [Mimblewimble] only has a single secret per [UTXO], the blinding factor (\\( k_i \\) ), the Tari protocol extends the [Mimblewimble] protocol to include scripting in the form of [TariScript]. This adds a second secret per [UTXO] called the script_key (\\( k_s \\) ). In practical terms this means that while vanilla [Mimblewimble] only requires that a wallet wishing to spend an [UTXO], prove knowledge of (\\( k_i \\) ) by singing the kernel signature, this is not sufficient for Tari. A Tari wallet must also prove knowledge of the script key (\\( k_s \\) ), by signing the script signature. 
+
+## Requirements
+
+To properly implement hardware wallets we need the following requirements to be met:
+* No UTXO can be spent without a user physically approving the transaction on the hardware wallet.
+* Users need to verify transaction properties when signing for transactions.
+* The user must be able to receive transactions without having to authorize them on the hardware wallet.
```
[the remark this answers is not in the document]
SWvheerden: Umm, with interactive payments, we cannot have a case of a `malicious sender effectively burning funds`. 
The receiver side if wholly constructed by the receiver. 
A malicious receiver is a whole different ball game and I don't think possible to stop

src/RFC-0205_LedgerTransactions.md
```diff
@@ -0,0 +1,148 @@
...
+All [UTXO]s created will be created with a script `PushPubkey(K_S)`. The key (\\( k_s \\) ) is created as follows:
+$$
+\begin{aligned}
+s = Hash(k_i * A) \\\\
+k_S = s + a \\\\
+K_S = s \cdot G + A
+\end{aligned}
+$$
+
+The blinding factor (\\( k_i \\) ) is used as a random nonce when creating the script key. This means the helper can create the public key without the signer present, and the signer can then at a later stage create the private key from the nonce. The key pair (\\( a, A \\) ) is the master private key from the signer. The private key (\\( a \\) ) is kept secret by the signer at all times.
```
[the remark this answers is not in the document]
SWvheerden: just made it key pair?

src/RFC-0205_LedgerTransactions.md
SWvheerden: Yeah this table is here for that regardless of how the key is leaked or learned

src/RFC-0205_LedgerTransactions.md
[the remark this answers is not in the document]
AaronFeickert: Isn't it still possible for a sender to duplicate an existing valid value-and-mask ciphertext with a malformed commitment? Then if the scanner is only defining "receiving" as being able to decrypt the ciphertext, it would be fooled and claim the recipient has funds that in fact cannot be spent.

src/RFC-0205_LedgerTransactions.md
```diff
@@ -0,0 +1,148 @@
...
+* Signer: This is the device that keeps the secrets for the transactions and approves them, aka the hardware wallet.
+* Helper: This device is the program that helps the signer construct the transaction and send it over the network, aka wallet.
+
+### Process Overview
+By splitting the ownership of the [UTXO]'s secret by assigning knowledge of only the script key (\\( k_s \\) ) to the signer, we can lift much of the heavy cryptography like bulletproof creation to the helper device by exposing (\\( k_i \\) ) to it. By looking at how [one-sided-stealth] transactions are created, we can construct the script key in such a way that the helper can calculate the public script key, but cannot calculate the private script key.
+
+All [UTXO]s created will be created with a script `PushPubkey(K_S)`. The key (\\( k_s \\) ) is created as follows:
+$$
+\begin{aligned}
+s = Hash(k_i * A) \\\\
```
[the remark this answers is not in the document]
AaronFeickert: This should be something akin to `H(k_i)`, not just `k_i` itself.

src/RFC-0205_LedgerTransactions.md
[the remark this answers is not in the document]
SWvheerden: The sender does not construct this, the receiver does. So how can the sender force the receiver to malform its own commitment?

src/RFC-0205_LedgerTransactions.md
AaronFeickert: Sure, just noting that you don't need `k_i` to do the brute forcing of the value, as `K_i` is sufficient for it too.

src/RFC-0205_LedgerTransactions.md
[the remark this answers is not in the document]
AaronFeickert: Looks like you already made an update later in the RFC to address this. Apologies!

src/RFC-0205_LedgerTransactions.md
```diff
@@ -0,0 +1,148 @@
...
+The helper can attach the correct signatures to the [UTXO]s and ship the transaction.
+
+### Receiving normal one-sided transaction
+This can be done by the helper asking the signer for a public key. And advertising this public key as the destination public key for a 1-sided transaction. 
+The helper can scan the blockchain for this public key. 
+
+### Receiving one-sided-stealth
+`TODO`
+
+### Output recovery
```
AaronFeickert, line 123: Perhaps worth noting that this process only works for interactive outputs.

For legacy one-sided outputs, the helper can watch for `A` in scripts (as you note elsewhere), but since it can't complete the DHKE required to decrypt the commitment opening, it can't assert the output is valid and must send it to the signer to be checked.

For stealth one-sided outputs, the helper can't do anything except send such outputs to the signer to be checked.

src/RFC-0205_LedgerTransactions.md
```diff
@@ -0,0 +1,148 @@
...
+
+All [UTXO]s created will be created with a script `PushPubkey(K_S)`. The key (\\( k_s \\) ) is created as follows:
+$$
+\begin{aligned}
+k_S = k_i + a \\\\
+K_S = k_i \cdot G + A
+\end{aligned}
+$$
+
+The blinding factor (\\( k_i \\) ) is used as a random nonce when creating the script key. This means the helper can create the public key without the signer present, and the signer can then at a later stage create the private key from the nonce. The key pair (\\( a, A \\) ) is the master key pair from the signer. The private key (\\( a \\) ) is kept secret by the signer at all times.
```
CjS77, line 99: When you say "master key pair", does this mean that the same `a` is used for every transaction?

src/RFC-0205_LedgerTransactions.md
```diff
@@ -0,0 +1,148 @@
...
+
+### Initialization
+Adding a hardware wallet to a wallet we need to ensure that all keys are only derived from a single seed phrase provided by the hardware wallet. 
+
+Helper asks signer for master helper key (\\( k_H \\) ). 
+This key is derived from the signer seed phrase.
+
+### Transaction receiving
+When a transaction is received the helper constructs the new [UTXO] with its Rangeproof. Choosing a new (\\( k_i \\) ) for the [UTXO], it calculates a new \\( s \\) and \\( K_S \\). It attaches the script `PushPubkey(K_S)` to output.
```
CjS77, line 108: What is `s` here?

src/RFC-0205_LedgerTransactions.md
```diff
@@ -0,0 +1,148 @@
...
+
+### Transaction sending
+When the user wants to send a transaction, the helper retrieves the desired [UTXO]. The helper gives asks the signer to sign the transaction.
```
CjS77, line 111:
```suggestion
When the user wants to send a transaction, the helper retrieves the desired [UTXO]. The helper asks the signer to sign the transaction. 
```

src/RFC-0205_LedgerTransactions.md
```diff
@@ -0,0 +1,148 @@
...
+
+### Entities
+Normal transactions have only a single entity, the wallet which controls all secrets and transactions. But with hardware wallets, we need to define two distinct types:
+* Signer: This is the entity that keeps the secrets for the transactions and approves them, aka the hardware wallet.
+* Helper: This entity is the program that helps the signer construct the transaction, send it over the network and scan the network, aka wallet.
+
+### Process Overview
+By splitting the ownership of the [UTXO]'s secret by assigning knowledge of only the script key (\\( k_s \\) ) to the signer, we can lift much of the heavy cryptography like bulletproof creation to the helper device by exposing (\\( k_i \\) ) to it. By looking at how [one-sided-stealth] transactions are created, we can construct the script key in such a way that the helper can calculate the public script key, but cannot calculate the private script key.
+
+All [UTXO]s created will be created with a script `PushPubkey(K_S)`. The key (\\( k_s \\) ) is created as follows:
```
CjS77, line 91: I gather, from the discussion below, that you mean to say that we must update the protocol to deprecate the current `NOP` default script so that there's always a script key present?

src/RFC-0205_LedgerTransactions.md
```diff
@@ -0,0 +1,148 @@
...
+### Receiving normal one-sided transaction
+This can be done by the helper asking the signer for a public key. And advertising this public key as the destination public key for a 1-sided transaction. 
+The helper can scan the blockchain for this public key. 
+
+### Receiving one-sided-stealth
+`TODO`
+
+### Output recovery
+When creating outputs the wallet encrypts the blinding factor (\\k_i \\) and value \\( v \\) with (\\( k_H \\) ).
+Because the key (\\( k_H \\) ) is calculated from the seed phrase of the signer, this will be the same each time. The helper can try to decrypt each scanned output, when it is successful it knows it has found its own output.
```
CjS77, line 125: Are `a` and `k_H` the same thing?

CjS77, line 125: So it's a bit worrying to use the same key every time from a security perspective.

Could we perhaps use a strategy similar to the one in [BIP-32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#user-content-Child_key_derivation_CKD_functions) for deriving an infinite number of `A'` from a single parent private key (signer) or the parent public key (helper)?

CjS77 (commented):
I like it. Some discussion points and clarifying questions.

CjS77 (commented):
I like it. Some discussion points and clarifying questions.

src/RFC-0205_LedgerTransactions.md
AaronFeickert: Yeah, for legacy one-sided it's mostly a UX issue, and somewhat a "wait for the signer" issue. For stealth one-sided it's an entirely different problem that does not scale.

src/RFC-0205_LedgerTransactions.md
AaronFeickert: Yes, it would be.

src/RFC-0205_LedgerTransactions.md
AaronFeickert: No, `k_H` is a fixed helper key derived from `a` that's used for encrypting commitment openings. The signer provides it once to the helper, who can then use it for AEAD operations.

AaronFeickert: There's no security problem here. Encrypted openings use an extended nonce whose entire purpose is to allow for safe key reuse with random nonce sampling. And this key reuse allows for super-fast interactive output scanning during recovery.

AaronFeickert: Also, the wallet public key `A` doesn't play any role in AEAD decryption during recovery. It only comes into play when testing for offset script public keys, and in that case, there's already a hash-based offset applied that masks the fixed `A`.

src/RFC-0205_LedgerTransactions.md
SWvheerden: was removed in a refactor, fixed

src/RFC-0205_LedgerTransactions.md
```diff
@@ -0,0 +1,148 @@
...
+
+### Transaction sending
+When the user wants to send a transaction, the helper retrieves the desired [UTXO]. The helper asks the signer to sign the transaction. 
+The singer calculates \\( k_s \\) to sign the transaction. 
+The signer creates a random nonce \\( k_O \\) to use for the script_offset. It produces the metadata signature with \\( k_O \\), and supplies the script_offset to the helper. 
+The helper can attach the correct signatures to the [UTXO]s and ship the transaction.
+
+### Receiving normal one-sided transaction
+This can be done by the helper asking the signer for a public key. And advertising this public key as the destination public key for a 1-sided transaction. 
+The helper can scan the blockchain for this public key.
```
stringhandler, line 118: What happens to the wallet's Tari Address? Is this not related?

src/RFC-0205_LedgerTransactions.md
```diff
@@ -0,0 +1,148 @@
...
+The singer calculates \\( k_s \\) to sign the transaction. 
+The signer creates a random nonce \\( k_O \\) to use for the script_offset. It produces the metadata signature with \\( k_O \\), and supplies the script_offset to the helper. 
+The helper can attach the correct signatures to the [UTXO]s and ship the transaction.
+
+### Receiving normal one-sided transaction
+This can be done by the helper asking the signer for a public key. And advertising this public key as the destination public key for a 1-sided transaction. 
+The helper can scan the blockchain for this public key. 
+
+### Receiving one-sided-stealth
+`TODO`
```
stringhandler, line 121: Just adding a marker here. DNM until resolved

stringhandler (commented):
I think it's worth adding a section on generic TariScripts that are not one-sided, e.g. HTLC

src/RFC-0205_LedgerTransactions.md
```diff
@@ -0,0 +1,148 @@
...
+the host machines that run the wallets are compromised by malware the secrets from the transactions are still safe as they are not
+kept on the machine as with regular wallets. The devices are very low power and only feature a very limited processing power. MimbleWimble 
+has intensive crypto operations that are in some instances very slow or not able to run on the hardware wallet at all. This RFC describes a way to 
+get around these limitations to have fully functional secure hardware wallet integration. 
+
+## Background
+
+Vanilla [Mimblewimble] only has a single secret per [UTXO], the blinding factor (\\( k_i \\) ), the Tari protocol extends the [Mimblewimble] protocol to include scripting in the form of [TariScript]. This adds a second secret per [UTXO] called the script_key (\\( k_s \\) ). In practical terms this means that while vanilla [Mimblewimble] only requires that a wallet wishing to spend an [UTXO], prove knowledge of (\\( k_i \\) ) by producing the kernel signature, this is not sufficient for Tari. A Tari wallet must also prove knowledge of the script key (\\( k_s \\) ), by producing the script signature.
```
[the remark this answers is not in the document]
SWvheerden: Practically No, Technically yes. 
A `No-Op` script that just contains the `No-Op` operation accepts any input. 

So from the wallet submitting the tx, it can submit any Public key as input, and it's still valid. But it still needs to prove it knows the secret key of that public key.  

But from the prover side, aka base_node, it just runs, the script and verifies the signature, thus proving knowledge of the script key k_S. How the key was calculated from the script the Base_node does not care.

src/RFC-0205_LedgerTransactions.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Background
+
+Vanilla [Mimblewimble] only has a single secret per [UTXO], the blinding factor (\\( k_i \\) ), the Tari protocol extends the [Mimblewimble] protocol to include scripting in the form of [TariScript]. This adds a second secret per [UTXO] called the script_key (\\( k_s \\) ). In practical terms this means that while vanilla [Mimblewimble] only requires that a wallet wishing to spend an [UTXO], prove knowledge of (\\( k_i \\) ) by producing the kernel signature, this is not sufficient for Tari. A Tari wallet must also prove knowledge of the script key (\\( k_s \\) ), by producing the script signature. 
+
+## Requirements
+
+To properly implement hardware wallets we need the following requirements to be met:
+* No UTXO can be spent without a user physically approving the transaction on the hardware wallet.
+* Users need to verify transaction properties when signing for transactions.
```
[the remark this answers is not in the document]
SWvheerden: yes

src/RFC-0205_LedgerTransactions.md
SWvheerden: Currently, the wallet calculates the Public script key for the script from the wallet's Tari Address. The only reason that is used is that it's an easily known key that the wallet has the secret key to. So we use that key in place of the TariAddress

src/RFC-0205_LedgerTransactions.md
AaronFeickert: It's faster, and there's no need to do ECDH for this use case.

src/RFC-0205_HardwareTransactions.md
```diff
@@ -0,0 +1,148 @@
...
+
+To properly implement hardware wallets we need the following requirements to be met:
+* No UTXO can be spent without a user physically approving the transaction on the hardware wallet.
+* Users need to verify transaction properties when signing for transactions.
+* The user must be able to receive transactions without having to authorize them on the hardware wallet.
+
+## Implementation
+
+### Entities
+Normal transactions have only a single entity, the wallet which controls all secrets and transactions. But with hardware wallets, we need to define two distinct types:
```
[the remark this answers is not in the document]
AaronFeickert: And I'd use the term "entities" rather than "types" here, for consistency.

src/RFC-0205_HardwareTransactions.md
```diff
@@ -0,0 +1,148 @@
...
+* Signer: This is the entity that keeps the secrets for the transactions and approves them, aka the hardware wallet.
+* Helper: This entity is the program that helps the signer construct the transaction, send it over the network and scan the network, aka wallet.
+
+### Process Overview
+By splitting the ownership of the [UTXO]'s secret by assigning knowledge of only the script key (\\( k_s \\) ) to the signer, we can lift much of the heavy cryptography like bulletproof creation to the helper device by exposing (\\( k_i \\) ) to it. By looking at how [one-sided-stealth] transactions are created, we can construct the script key in such a way that the helper can calculate the public script key, but cannot calculate the private script key.
```
AaronFeickert, line 89: Some context on what a UTXO's "secret" is could be helpful here. Really it's two secret values: the commitment mask and the script private key. (The value is private too, but can be easily brute forced and therefore not as relevant.)

src/RFC-0205_HardwareTransactions.md
```diff
@@ -0,0 +1,148 @@
...
+
+All hardware wallet created [UTXO]s created will be created with a script `PushPubkey(K_S)`. The key (\\( k_s \\) ) is created as follows:
+$$
+\begin{align}
+k_S &= k_i + a \\\\
```
AaronFeickert, line 94: Still recommend not using the mask directly in this way.

src/RFC-0205_HardwareTransactions.md
```diff
@@ -0,0 +1,148 @@
...
+K_S &= k_i \cdot G + A
```
AaronFeickert, line 95: Nor here.

src/RFC-0205_HardwareTransactions.md
AaronFeickert: If this is for interactive outputs, I'd use an approach like `H(k_i) + a`. If the design choice is made to use the helper key `k_H` for the commitment opening encryption, the helper won't need to compute the corresponding script public key `H(k_i)*G + A` except on outputs it already flags via opening decryption.

src/RFC-0205_HardwareTransactions.md
SWvheerden: It is discussed two paragraphs up?

src/RFC-0205_HardwareTransactions.md
SWvheerden: Its going to have to compute `K_S` for every commitment it makes, it needs to put that into the script

AaronFeickert: Right; I meant for scanning purposes (but wasn't clear about that).

AaronFeickert: A similar attack could be used in an attempt to generically link transactions without knowledge of `A`.

AaronFeickert (changes requested):
There are a few spots where the math-mode separators are written incorrectly, meaning not all math renders properly.

src/RFC-0205_HardwareTransactions.md
```diff
@@ -0,0 +1,147 @@
...
+the host machines that run the wallets are compromised by malware the secrets from the transactions are still safe as they are not
+kept on the machine as with regular wallets. The devices are very low power and only feature a very limited processing power. MimbleWimble 
+has intensive crypto operations that are in some instances very slow or not able to run on the hardware wallet at all. This RFC describes a way to 
+get around these limitations to have fully functional secure hardware wallet integration. 
+
+## Background
+
+Vanilla [Mimblewimble] only has a single secret per [UTXO], the blinding factor (\\( k_i \\) ), the Tari protocol extends the [Mimblewimble] protocol to include scripting in the form of [TariScript]. This adds a second secret per [UTXO] called the script_key (\\( k_s \\) ). In practical terms this means that while vanilla [Mimblewimble] only requires that a wallet wishing to spend an [UTXO], prove knowledge of (\\( k_i \\) ) by producing the kernel signature, this is not sufficient for Tari. A Tari wallet must also prove knowledge of the script key (\\( k_s \\) ), by producing the script signature.
```
AaronFeickert, line 71: The first sentence is a run-on.

src/RFC-0205_HardwareTransactions.md
```diff
@@ -0,0 +1,147 @@
...
+The signer calculates \\( k_s \\) to sign the transaction. 
+The signer creates a random nonce \\( k_O \\) to use for the script_offset. It produces the metadata signature with \\( k_O \\), and supplies the script_offset to the helper. 
+The helper can attach the correct signatures to the [UTXO]s and ship the transaction.
+
+### Receiving normal one-sided transaction
+This can be done by the helper asking the signer for a public key. And advertising this public key as the destination public key for a 1-sided transaction. 
+The helper can scan the blockchain for this public key. 
+
+### Receiving one-sided-stealth
+`TODO`
```
AaronFeickert, line 120: Not possible yet with this design.

src/RFC-0205_HardwareTransactions.md
```diff
@@ -0,0 +1,147 @@
...
+
+### Output recovery
+When creating outputs the wallet encrypts the blinding factor \\(k_i \\) and value \\( v \\) with \\( k_H \\).
```
AaronFeickert, line 123: Might be worth noting that the encryption is done in practice using an extended-nonce AEAD, which allows for a random nonce and uses authenticated decryption for this detection process.

AaronFeickert (changes requested):

src/RFC-0205_HardwareTransactions.md
```diff
@@ -0,0 +1,147 @@
...
+
+### Entities
+Normal transactions have only a single entity, the wallet which controls all secrets and transactions. But with hardware wallets, we need to define two distinct entities:
+* Signer: This is the entity that keeps the secrets for the transactions and approves them, aka the hardware wallet.
+* Helper: This entity is the program that helps the signer construct the transaction, send it over the network and scan the network, aka wallet.
+
+### Process Overview
+By splitting the ownership of the [UTXO]'s secrets by assigning knowledge of only the script key (\\( k_s \\) ) to the signer, we can lift much of the heavy cryptography like bulletproof creation to the helper device by exposing (\\( k_i \\) ) to it. By looking at how [one-sided-stealth] transactions are created, we can construct the script key in such a way that the helper can calculate the public script key, but cannot calculate the private script key.
+
+All hardware wallet created [UTXO]s created will be created with a script `PushPubkey(K_S)`. The key (\\( k_s \\) ) is created as follows:
```
brianp, line 90: That's a lot of created.

```suggestion
All hardware wallet created [UTXO]s will be contain a script `PushPubkey(K_S)`. The key (\\( k_s \\) ) is created as follows:
```

CjS77 (approved):

PR #100, opened 2023-06-06
feat: rfc-0154 contacts and profile deep links

TruszczynskiA:
- Added /contacts and /profile deep links descriptions to the RFC-0154


SWvheerden (approved):
Ack

CjS77 (approved):

PR #103, opened 2023-07-31
feat: range proof performance and notation fixes

### LINKED ISSUE #102: Add comparative performance data to range proof RFC
Discussion in #101 included in a suggestion to include comparative performance data in range proof [RFC-0181](https://rfc.tari.com/RFC-0181_BulletproofsPlus.html). That PR was merged without this addition in order to better assist reviewers in an ongoing audit, but it would still be useful to include.

AaronFeickert:
Description
---
Updates range proof [RFC-0181](https://rfc.tari.com/RFC-0181_BulletproofsPlus.html) to add comparative performance data and improve notation.

Closes #102.

Motivation and Context
---
An [earlier PR](https://github.com/tari-project/rfcs/pull/101) removed outdated benchmark data, but it was [suggested](https://github.com/tari-project/rfcs/pull/101#issuecomment-1646739122) to include reasonable comparative performance data against Bulletproofs. This data has been added for both proof size and verification complexity.

Some notation was incomplete or incorrect. In particular, sum and index bounds were left implicit in many places, and indexing relating to mask recovery was incorrect. Both of these have been addressed.

How Has This Been Tested?
---
The changes build and appear to render correctly.

### DIFF
--- a/src/RFC-0181_BulletproofsPlus.md
+++ b/src/RFC-0181_BulletproofsPlus.md
@@ -74,9 +74,12 @@ An approach to mask and value recovery was [used by Grin](https://github.com/mim
 
 ## Notation
 
-To reduce confusion in our description and more closely match implementation libraries, we use additive notation and uppercase letters for group elements, and otherwise assume notation from the preprints.
+To reduce confusion in our description and more closely match implementation libraries, we use additive notation and uppercase letters for group elements, and otherwise generally assume notation from the preprints.
 Denote the commitment value generator by $G\_c$ and the commitment mask generator vector by $\vec{H}\_c$.
 Because the preprint uses the notation $A$ differently in the weighted inner product and range proving protocols, we rename it to $A'$ in the weighted inner product protocol.
+
+We assume that a range proof aggregates $m$ range assertions, each of which is to an $n$-bit range.
+We further assume that each value commitment uses $p$ masks; that is, a standard Pedersen commitment would have $p = 1$.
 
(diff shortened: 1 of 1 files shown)


CjS77 (approved):
LGTM

hansieodendaal (commented):
@AaronFeickert, thanks for adding the performance discussion and other refinements. 

The only outstanding question is the relative proving time between BP and BP+. Initial benchmarks showed that BP+ is 30% more expensive on average. I quote the previous version of the RFC:

```
#### BP vs. BP+ (creation)

BP+ creation is 30% slower than BP.

| Size | BP Med (ms) | BP+ Med (ms) | Diff Med (%) |
|------|-------------|--------------|--------------|
| 1    | 16.29       | 21.24        | 130%         |
| 2    | 31.63       | 41.08        | 130%         |
| 4    | 60.47       | 80.46        | 133%         |
| 8    | 119.18      | 156.56       | 131%         |
| 16   | 240.18      | 306.03       | 127%         |
| 32   | 460.67      | 598.57       | 130%         |
|      |             | Average      | 130%         |
```

AaronFeickert:
Should proving complexity be considered important enough to include? They're often given much less weight since they're typically one-off operations whose complexity is irrelevant to verifiers. A one-time increase of a few milliseconds (to a prover) doesn't matter in the same way that a million increases of a few milliseconds (to a verifier) would be.

hansieodendaal:
I think it is important to add something, yes. It can always be put into context, as you pointed out. I would imagine that on some devices proving time would matter, like on a mobile phone when transactions are created.

PR #105, opened 2023-08-01
docs: sparse merkle trees

CjS77:
This Request for Comment (RFC) proposes replacing the current Mutable Merkle Mountain Range (MMMR) data structure 
used for tracking the commitment to the UTXO set, with a Sparse Merkle tree (SMT).

A sparse Merkle tree ([SMT]) is a Merkle-type structure, except the contained data is indexed, and each datapoint is 
placed at the leaf that corresponds to that datapoint’s index. Empty nodes are represented by a predefined "null" value.

Since every empty node has the same hash, and the tree is sparse, it is possible to prune the tree in a way such 
that only the non-zero leaf nodes and placeholders marking the empty sub-trees are stored. Therefore, SMTs are 
relatively compact.

### DIFF
--- a/src/RFC-0141_Sparse_Merkle_Trees.md
+++ b/src/RFC-0141_Sparse_Merkle_Trees.md
@@ -69,7 +69,7 @@ relatively compact.
 A major feature of SMTs is that they are truly mutable. The current UTXO Merkle root in Tari is calculated using a 
 Merkle Mountain Range (MMR). This has some drawbacks: 
 
-1. MMRs are _immutable_ data structures,and therefore as a workaround (some would say, hack), a bitmap is appended 
+1. MMRs are _immutable_ data structures, and therefore as a workaround (some would say, hack), a bitmap is appended 
    to the MMR to mark the spent outputs. In Tari's implementation, a roaring bitmap is used, which takes 
    advantage of compression, but even so, it is still fairly large and will grow indefinitely.
 2. The Merkle tree must keep a record of all TXOs forever, and mark them as they are spent. The blockchain cannot 
@@ -83,7 +83,7 @@ SMTs are true mutable data structures and do not have these drawbacks.
 
 1. No tracking bitmap is needed. When a UTXO is spent, it can be deleted from the tree.
 2. It is possible to prune STXOs from the UTXO set to calculate the Merkle root.
-3. Adding and removing UTXOs in any order will always yield the same Merkle root. Adding, and then deleting a UTXO 
+3. Adding and removing UTXOs in any order will always yield the same Merkle root. Adding and then deleting a UTXO 
    from the set will result in the same Merkle root as before the UTXO was added.
 
 ### Inclusion and exclusion proofs
@@ -109,15 +109,15 @@ Consider some representative numbers:
 Let's assume there are 1,000,000 UTXOs, with another 2,000,000 UTXOs having being spent over the lifetime of the
 project. A busy blockchain might achieve this level of traffic in a few days.
 
-If each commitment-UTXO hash pair is 64 bytes, you need serialize 64MB to recreate the merkle root for the SMT.
+If each commitment-UTXO hash pair is 64 bytes, you need serialize 64MB to recreate the Merkle root for the SMT.
 
-For the MMMR, even though you only need the UTXO hash, you need all 3,000,0000 values (96MB) plus approximately 1MB 
+For the MMMR, even though you only need the UTXO hash, you need all 3,000,000 values (96MB) plus approximately 1MB 
 for every million hashes in a bitmap to indicate which hashes have  been deleted (3 MB) for a total of 99MB.
 
 This only gets worse with time. Over a period of a year, a busy blockchain might have 100,000,000 spent transaction 
 outputs. However, the UTXO set will grow far more slowly, and perhaps only 10x in size to 10 million outputs.
 
-The SMT tree requires serialising 640MB of data to recreate the root, whereas the MMMR now requires 3.6GB of data.
+The SMT requires serialising 640MB of data to recreate the root, whereas the MMMR now requires 3.6GB of data.
 
 ## Implementation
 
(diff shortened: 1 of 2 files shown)


CjS77:
ACK

SWvheerden (commented):
ACK

src/RFC-0141_Sparse_Merkle_Trees.md
```diff
@@ -0,0 +1,669 @@
...
+Let's create a SMT with four nodes.
+
+If we insert the nodes at
+ * A: 01001111 (79 in decimal)
+ * B: 01011111 (95 in decimal)
+ * C: 11100000 (224 in decimal)
+ * D: 11110000 (240 in decimal)
+
+you will notice that the first two diverge at the first bit, while the first and last pairs differ at the
+fourth bit. This results in a SMT that looks like this:
```
AaronFeickert, line 164: The first and second indexes differ at the fourth bit, whereas the first and last pairs differ at the first bit.

src/RFC-0141_Sparse_Merkle_Trees.md
```diff
@@ -0,0 +1,669 @@
...
+```text
+    H::digest(BRANCH_PREFIX || height || key_prefix || left_child_hash || right_child_hash)
+```
+
+where
+
+* `height` is the height of the branch in the tree, where the root node is height 0.
+* `key_prefix` is the common prefix that the key of _every_ descendent node of this branch will begin with. 
+  `key_prefix` is `KEY_LENGTH_BYTES` long, and every bit after the prefix MUST be set to zero. This means that key 
+  prefixes are not unique. For example, every key prefix for the left-mode path down the tree will always have a
```
AaronFeickert, line 345: Do you mean "left-most" here?

src/RFC-0141_Sparse_Merkle_Trees.md
```diff
@@ -0,0 +1,669 @@
...
+        // calculate expected leaf node hash
+        let leaf_hash = LeafNode::<H>::hash_value(expected_key, expected_value);
+        let calculated_root = self.calculate_root_hash(expected_key, leaf_hash);
+        calculated_root == *expected_root
+    }
+```
+
+## Backwards Compatibility
+
+If the UTXO merkle root is replaced by a sparse merkle tree, this change would require a **hard fork**,
```
AaronFeickert, line 654:
```suggestion
If the UTXO Merkle root is replaced by a sparse Merkle tree, this change would require a **hard fork**, 
```

src/RFC-0141_Sparse_Merkle_Trees.md
```diff
@@ -0,0 +1,669 @@
...
+        let leaf_hash = LeafNode::<H>::hash_value(expected_key, expected_value);
+        let calculated_root = self.calculate_root_hash(expected_key, leaf_hash);
+        calculated_root == *expected_root
+    }
+```
+
+## Backwards Compatibility
+
+If the UTXO merkle root is replaced by a sparse merkle tree, this change would require a **hard fork**, 
+since it fundamentally alters how the UTXO merkle root is calculated.
```
AaronFeickert, line 655:
```suggestion
since it fundamentally alters how the UTXO Merkle root is calculated. 
```

AaronFeickert (changes requested):

src/RFC-0141_Sparse_Merkle_Trees.md
```diff
@@ -0,0 +1,669 @@
...
+Let's create a SMT with four nodes.
+
+If we insert the nodes at
+ * A: 01001111 (79 in decimal)
+ * B: 01011111 (95 in decimal)
+ * C: 11100000 (224 in decimal)
+ * D: 11110000 (240 in decimal)
+
+you will notice that the first two diverge at the first bit, while the first and last pairs differ at the
+fourth bit. This results in a SMT that looks like this:
```
CjS77, line 164:
```suggestion
you will notice that the first two diverge at the fourth bit, while the first and last pairs differ at the
first bit. This results in a SMT that looks like this:
```

src/RFC-0141_Sparse_Merkle_Trees.md
```diff
@@ -0,0 +1,669 @@
...
+   
+<div style="line-height: 1em" class="hljs">
+<pre>
+            ┌──────┐
+      ┌─────┤ root ├─────┐
+      │     └──────┘     │
+     ┌┴┐                ┌┴┐1
```
CjS77, line 171:
```suggestion
     ┌┴┐0               ┌┴┐1
```

src/RFC-0141_Sparse_Merkle_Trees.md
```diff
@@ -0,0 +1,669 @@
...
+  ┌──┤ ├──┐          ┌──┤ ├───┐
+  │  └─┘  │          │  └─┘   │
+ ┌┴┐     ┌┴┐        ┌┴┐10    ┌┴┐11
```
CjS77, line 174:
```suggestion
 ┌┴┐00   ┌┴┐01      ┌┴┐10    ┌┴┐11
```

src/RFC-0141_Sparse_Merkle_Trees.md
```diff
@@ -0,0 +1,669 @@
...
+ │0│  ┌──┤ ├──┐     │0│    ┌─┤ ├─┐
+ └─┘  │  └─┘  │     └─┘    │ └─┘ │
+     ┌┴┐     ┌┴┐       110┌┴┐   ┌┴┐111
```
CjS77, line 177:
```suggestion
     ┌┴┐010  ┌┴┐011     110┌┴┐   ┌┴┐111
```

src/RFC-0141_Sparse_Merkle_Trees.md
```diff
@@ -0,0 +1,669 @@
...
+```text
+    H::digest(BRANCH_PREFIX || height || key_prefix || left_child_hash || right_child_hash)
+```
+
+where
+
+* `height` is the height of the branch in the tree, where the root node is height 0.
+* `key_prefix` is the common prefix that the key of _every_ descendent node of this branch will begin with. 
+  `key_prefix` is `KEY_LENGTH_BYTES` long, and every bit after the prefix MUST be set to zero. This means that key 
+  prefixes are not unique. For example, every key prefix for the left-mode path down the tree will always have a
```
CjS77, line 345:
```suggestion
  prefixes are not unique. For example, every key prefix for the left-most path down the tree will always have a 
```

AaronFeickert (approved):

PR #112, opened 2023-11-20
feat: add rfc 321

stringhandler:
Adds an RFC for handling foreign DAN proposals


src/RFC-0321_ProcessingForeignProposals.md
```diff
@@ -0,0 +1,113 @@
...
+
+NOTE: ForeignProposal commands can be proposed in between a previous ForeignProposal and LocalPrepare/Timeout commands, but commands from the ForeignProposal **must** only be proposed after all transactions in the first ForeignProposal have been sequenced. 
+The TIMEOUT_TIME block is counted from the height where the ForeignProposal is sequenced.
+
+ForeignProposal commands **must** appear in strict ascending order in the blockchain, but do not have to be in sequential blocks. In other words, for shard *s*, the block containing ForeignProposal(*s*, 1) must have a height lower than ForeignProposal(*s*, 2). Also, if a chain contains ForeignProposal(*s*, 1) and ForeignProposal(*s*, 3), then it **must** also contain ForeignProposal(*s*, 2).
+
+If a node receives a foreign proposal (not the command), and it has not received
+the previous foreign proposal, then it should ask the committee to provide it to them.
+
+If the LOCAL_PREPARE command appears in a local proposal for a transaction and is the last such LOCAL_PREPARE and all shards have voted to accept, an ALL_PREPARED(tx) command **must** also be present in the local proposal.
```
Cifko, line 103: So if we have LOCAL_PREPARE(tx) and we have all foreign LOCAL_PREPARE already, we should put ALL_PREPARED(tx) instead? Or do you mean both will be present? What will happen when the block is too big and it's not enough space to add both?

src/RFC-0321_ProcessingForeignProposals.md
```diff
@@ -0,0 +1,113 @@
...
+To solve the above problems, we'll use reliable broadcast and process foreign evidence in order.
+
+In a local shard committee, the proposed block **must** include a counter per epoch and per shard. If the proposal includes transactions that involve other shards, this counter **must** be incremented.
+
+When the proposed block becomes committed locally, the block **must** be broadcast to each involved shard that was incremented, along with evidence of being committeed (The chain of QCs must be included).
+
+To ensure this, each node in the local committee will forward this committed block, along with a 3 chain of QC's proving it was committeed.
+
+As a local committee member, when I receive a foreign proposal, if it is valid I will queue up a special command ForeignProposal(number, QC_Hash) that I must propose 
+when I am next leader (if it has not been proposed already). I also **should** request all transaction hashes that I have not seen from involved_shards for each transaction in the proposal, and add them to my mempool for execution.
```
Cifko, line 82: This is just for notifying the rest of the committee that they should be aware of the foreign proposal? They will need it anyway when they will try to validate TX going from `local_prepared` to `accept`. Voting on this special command is the same as voting on individual transactions that includes foreign shards. Or am I missing something?

src/RFC-0321_ProcessingForeignProposals.md
```diff
@@ -0,0 +1,113 @@
...
+
+In the same order of transactions in the proposal, the transactions must be sequenced into the local chain as either a TIMEOUT or a LOCALPREPARE(TxId, ForeignShardId).
```
Cifko, line 84: I think the ordering of transactions within a proposal will cause some inconsistency.
E.g. I have Shard Sa and Sb, and Sa receive Ta (transaction over both shards) and Sb receive Tb (also over both shards). Then Sa will prepare a block with ordering Ta,Tb and Sb will prepare a block with order Tb,Ta. Then the foreign proposal will have wrong order compared to local proposal.

Why does the order in the proposal matters? It's not like I can have concurrent transactions in the proposal, right?

src/RFC-0321_ProcessingForeignProposals.md
stringhandler: The main reason is to set a starting point for a timeout. If the ForeignProposal command is committed, then `2f+1` nodes have agreed that this data has arrived and all of them will try to retrieve the transaction. If the transaction does exist, then at least one of them should find it and be able to propose it, if none of the nodes can find it in that time, the transaction must timeout.

src/RFC-0321_ProcessingForeignProposals.md
stringhandler: I suppose the order inside the proposal does not matter, so long as the proposals themselves are processed in entirety before any transactions for a new proposal can be sequenced. The rule should then be updated that any transactions that have not been processed within the time limit must be timed out

src/RFC-0321_ProcessingForeignProposals.md
stringhandler: Maybe we can just put the ALL_PREPARED in then, and same for SOME_PREPARED and ALREADY_ABORTED

src/RFC-0321_ProcessingForeignProposals.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Goals
+
+This RFC describes the process of distributing and processing foreign proposals in the Tari DAN Cerberus Model
+
+Across the entire network transactions must be processed or time out. When a transaction is started on a shard, it locks up substates, preventing other transactions from completing. Therefore if a transaction is started on a shard, it 
+should complete or be aborted in a timely manner to release the resources.
+
+Ordering of transactions is also important across all shards. While the different
+shards form a Direct Acyclic Graph (DAG), there should be a consistent ordering
```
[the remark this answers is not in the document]
stringhandler: Given the later comments, perhaps I should remove this requirement and add it back later if we need it

stringhandler:
@Cifko made some changes, please have a look

src/RFC-0321_ProcessingForeignProposals.md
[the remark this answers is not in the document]
Cifko: sounds good

src/RFC-0321_ProcessingForeignProposals.md
Cifko: I was expecting that the timer starts once I send something to the foreign committees and I expect an answer. E.g. I send Prepare(tx)/LocalPrepare to foreign committee, I start a timer that I'm expecting to receive something.

src/RFC-0321_ProcessingForeignProposals.md
Cifko: I'm curious when can it happen that I receive a foreign proposal with some Tx and I'm not able to do localprepare(tx). That means that even the committee from which I got the foreign proposal don't have the Tx.
Now I understand why you want timeout in foreign committee to send you that it timed out.

src/RFC-0321_ProcessingForeignProposals.md
```diff
@@ -0,0 +1,108 @@
...
+
+As a local committee member, when I receive a foreign proposal, if it is valid I will queue up a special command ForeignProposal(number, QC_Hash) that I must propose 
+when I am next leader (if it has not been proposed already). I also **should** request all transaction hashes that I have not seen from involved_shards for each transaction in the proposal, and add them to my mempool for execution.
+
+Before transactions in the `N+1`th foreign proposal for a shard, all transactions in the `N`th foreign proposal for that shard must be sequenced into the local chain as either a TIMEOUT or a LOCALPREPARE(TxId, ForeignShardId).
+
+When a LOCALPREPARE command is found in a local proposal, there **must** be a 
+Prepare(tx) command for that transaction in or before the current proposal.
+
+In the `TIMEOUT_TIME`` block  (e.g. 1000 blocks, but should probably be equal to committee_size * X rounds) after the ForeignProposal command has been proposed,
```
stringhandler, line 84: Add "Dummy blocks should be excluded from this count"

sdbondi (commented):
Thanks for writing this up. It's a complex subject. 

I agree that we need a way to ABORT/TIMEOUT a transaction if a LocalPrepare is never received. If a node receives `AllPrepared`/`SomePrepared` and does not have all foreign shard QCs listed in the evidence (or perhaps entire the 3-chain evidence, I think the current Evidence struct is too simplistic), it could request them from the proposer before voting. I think this is equivalent to agreeing on "having the foreign proposal". 

As always, there's a ton of ways to approach this and maybe the best way is to attempt an implementation and see what realisations we have.

src/RFC-0321_ProcessingForeignProposals.md
```diff
@@ -0,0 +1,108 @@
...
+
+To ensure this, each node in the local committee will forward this committed block, along with a 3 chain of QC's proving it was committeed.
+
+As a local committee member, when I receive a foreign proposal, if it is valid I will queue up a special command ForeignProposal(number, QC_Hash) that I must propose 
+when I am next leader (if it has not been proposed already). I also **should** request all transaction hashes that I have not seen from involved_shards for each transaction in the proposal, and add them to my mempool for execution.
+
+Before transactions in the `N+1`th foreign proposal for a shard, all transactions in the `N`th foreign proposal for that shard must be sequenced into the local chain as either a TIMEOUT or a LOCALPREPARE(TxId, ForeignShardId).
+
+When a LOCALPREPARE command is found in a local proposal, there **must** be a 
+Prepare(tx) command for that transaction in or before the current proposal.
```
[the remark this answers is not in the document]
stringhandler: The main difference here is to make sure we start the transaction locally before committing to other shards

src/RFC-0321_ProcessingForeignProposals.md
```diff
@@ -0,0 +1,108 @@
...
+
+In the `TIMEOUT_TIME`` block  (e.g. 1000 blocks, but should probably be equal to committee_size * X rounds) after the ForeignProposal command has been proposed, 
+all transactions in the proposal must be sequenced, either as LOCALPREPAREs or TIMEOUTs. This is to cater for the case where a transaction was prepared by a foreign committee, but was never found on the network. This is unlikely to happen 
+in a non-malicious scenario. The more common case is that all transactions will be found and have a LOCALPREPARE, but a block at `TIMEOUT_TIME` **must** be considered invalid if there are any transactions that are not LOCALPREPARED or TIMEOUT from the foreign proposal that started the timeout.
```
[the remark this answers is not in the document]
stringhandler: Because of the new Command ForeignProposal (or perhaps ForeignProposalReceived is a better name), the nodes in the local committee have all agreed that they have received the foreign proposal, so if this command is in the local chain, the majority acknowledge that they have the ForeignProposal in hand

src/RFC-0321_ProcessingForeignProposals.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Description
+
+To solve the above problems, we'll use reliable broadcast and process foreign evidence in order.
+
+In a local shard committee, the proposed block **must** include a counter per epoch and per shard (except for the local shard). If the proposal includes transactions that involve other shards, this counter **must** be incremented.
+
+When the proposed block becomes committed locally, the block **must** be broadcast to each involved shard that was incremented, along with evidence of being committeed (The chain of QCs must be included).
+
+To ensure this, each node in the local committee will forward this committed block, along with a 3 chain of QC's proving it was committeed.
```
Cifko, line 74: Well, if you send only 3 QCs you can't prove they are actually from one 3 chain. You can prove that you have a QC1 on top of the fLP block. But you can't prove that you have a QC2 on top of the block with QC1.
Maybe we should add block.justify() not just block_id to the QC. But I'm hope that we can come up with a better solution.

PR #116, opened 2023-12-05
chore: add epoch management to summary

CjS77:
- Added RFC-235 to summary, so it appears in the list.
- Editing for grammar, concision and clarity.


src/RFC-0325_DanTimeManagement.md
```diff
@@ -0,0 +1,188 @@
...
+before that it is only `VN_1`. From this it can be seen that `VN_1` thinks only it must be in the [VNC] but `VN_2`
+thinks they both need to be. But using the DGT of 1 block, `VN_1` will accept `VN_2` as part of the [VNC]
+because it is withing the DGT period and it assumes that its own [base node] is simply lagging 1 block.
+
+The DGT has the additional benefit of allowing a [VN] to finish processing an instruction after a block height has been
+reached.
+
+With the example mentioned above if `VN_1` has to stop processing an instruction at height 1000, but it started the
+instruction at height 999. When height 1000 comes
+along it can still continue to process the instruction till height 1001.
```
CjS77, line 114: I feel that a more effective solution to this problem is to treat this as a consensus issue on the DAN layer.

A leader can propose `NEW_EPOCH` to the presumed new VNC. If less than 66% of VNs have updated their epoch, the motion fails, and the VNC cannot form.

We should do a rough safety and liveness proof of whether this works based on:
- an honest node does not equivocate on its epoch. Once a `NEW_EPOCH` is accepted, a VN will ignore messages from their old VNC/epoch combination.
- A VNC must respond to and accept valid proposals from the old VNC/epoch combo until a new epoch has been accepted.

One consequence of this is that we may need to shuffle 33%-66% of nodes every epoch so that consensus and a new committee lines up in terms of numbers.

PR #117, opened 2023-12-11
Correction

lwillacy:
Description
Fixing an incorrect reference to the list of conditions validator nodes must verify to confirm stablecoin transactions.  The list of conditions is under section 2, but the reference is to section 3.

Motivation and Context
N/A

How Has This Been Tested?
N/A


AaronFeickert:
Superseded by #121, which makes additional corrections.

PR #118, opened 2023-12-12
feat: cerberus detail

CjS77:
First draft of PCerb description with transaction flow, and accompanying  mermaid diagram.

 The way that substate addresses (currently called ShardIds in the code) are derived is fairly elaborate. This commit elucidates the concept. It  largely interprets how the code has been written.


CjS77:
RFC still in draft state, but merging it for now so that kapa.ai can pick up the content

PR #131, opened 2024-03-06
feat: node-based build script

CjS77:
Add a node.js-based build script so that this site can run in Cloudflare pages.

The script does the following:
1. Downloads the specified version of mdbook from the GitHub releases page
2. Verifies its SHA256 hash
3. Extracts it to a temporary directory
4. Runs mdbook build against the book.toml file in the current directory
5. Places the built book in the temporary directory

`mdbook.env` (which is checked in and MUST NOT hold any secrets) allows you to set the version of mdbook and its hash.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 target
 .env
+build
+node_modules
 
 # Ignore IDE workspace files
 .idea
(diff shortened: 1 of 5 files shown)


CjS77:
ACK

PR #135, opened 2024-03-21
Tari Universe communication with provider

MCozhusheck:
This RFC description how communication between tapplet and wallet daemon could look like in Tari Universe.

Communication between web app and wallet daemon normally requires webRTC connection. In out case this not necessary as both wallet daemon and tapplet are on same host machine. This requires another way of communication between tapplet and wallet daemon.


stringhandler (commented):
There is already a client JS library for interacting via websockets and Tari's Metamask Snap (https://github.com/tari-project/tari.js). Is there a way the developer can include a single library that works with all three?

PR #136, opened 2024-03-21
chore: remove unintentional double negative

AaronFeickert:
Description
---
This PR doesn't not remove an unintended double negative.

Motivation and Context
---
The presence of the double negative doesn't not change the description of scripting support across the ecosystem.

How Has This Been Tested?
---
The page renders not incorrectly, and doesn't not fail not to build.


CjS77 (approved):
Not unapproved :)

PR #137, opened 2024-03-21
docs: rfc-1101 tapplet

karczuRF:
The aim of this Request for Comment (RFC) is to propose the Tapplet concept. Tapplets are dedicated Tari's applets registered on the L1 to ensure a superior level of security for network users.


stringhandler (commented):
Looking good so far. We need to have a signature to make sure that the contents have not been changed

src/RFC-1101_Tapplet.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Related Requests for Comment
+
+[RFC-1100](https://github.com/tari-project/rfcs/pull/134)
+
+## Description
+
+### Motivation
+
+The **Tapplet** is dedicated type of applet for the Tari blockchain and the fundamental part of the Tari Universe described in [RFC-1100](https://github.com/tari-project/rfcs/pull/134). The idea behind registered tapplets is to solve the class of vulnerabilities of the Dapp model known as supply chain attacks. This covers a wide range of attacks like: injecting malware code, XSS, stealing domains, social engineering attacks, etc.
```
CjS77, line 61: Todo - update link after RFC 1100 is merged.

src/RFC-1101_Tapplet.md
```diff
@@ -0,0 +1,123 @@
...
+The solution proposed by Tari uses the tapplets Registry (another RFC will describe it) to register hashes of zip bundles of applications and sign them using the private key associated with Yat. The actual zip bundle can be hosted by https, but thanks to the registry commitment, the client can verify that the bundle hasn’t been tampered with.
```
CjS77, line 62: TODO `the tapplets Registry (another RFC will describe it)` => `[Tapplets Registry](link/to/rfc/)`

src/RFC-1101_Tapplet.md
```diff
@@ -0,0 +1,123 @@
...
+Additionally the idea is to create an on-chain and association between smart contracts and their front end tapplets. This way, when a smart contract gets reused for a different purpose, it already comes with a ready tapplet that can be nested into other tapplets.
+
+### Tapplets
+
+At first, it is proposed that tapplets are npm packages published to the public npm registry, but in the future it may be extended also to other package managers. Npm is the first choice because it is widely used, so publishing a new Tapplet is as simple as publishing an npm package.
```
CjS77, line 67: Can we think of an alternative? You points are fair, but npm has such a dismal security record that it might hurt the perceived integrity of the product.

Off the top of my head, I can't think of an alternative, besides IPFS, obviously.

Otherwsie maybe we can add armour to the npm packages, e.g. by utilitsing elements of https://theupdateframework.com ?

src/RFC-1101_Tapplet.md
```diff
@@ -0,0 +1,123 @@
...
+
+Tapplets are separate packages and the file structure depends on its authors, however every package should contain a helpful readme file and must contain a manifest json file with Tapplet data.
+The following details are specific to tapplets:
+
+- The *version* string field in `package.json` and `tapplet.manifest.json` should match.
```
CjS77, line 72: Name of, and
Public key of publisher?

src/RFC-1101_Tapplet.md
```diff
@@ -0,0 +1,123 @@
...
+        "packageName": "@company-name/tapplet-name",
+        "registry": "https://registry.npmjs.org/"
+      }
+    }
+  },
+  "manifestVersion": "1.2.3"
+}
+
+```
+
```
CjS77, line 118: Some questions that arise that aren't clear from the current text:

* How do you propose managing upgrades? It's a complex topic on it's own to be sure.
* Where does the manifest live / Are there commitments to it anywhere?
* The manifest is mutable. How do you handle attempts to change the content hash (to point to a malicious app) without changing the version? (If it's meant to be immutable, the status field isn't useful.)

CjS77 (commented):
Good start, but needs a bit more detail to understand the proposal completely.

src/RFC-1101_Tapplet.md
karczuRF: > npm has such a dismal security record 

Could you please elaborate on what you mean by that? 
This is really good point and I'm gonna include this in the RFC, just need more context to compare with (if any) alternatives. 

TUF might be the way to go!

src/RFC-1101_Tapplet.md
karczuRF: These are really good points! I'll described them detailed in the RFC document, here just briefly. 
1. Each version of a tapplet needs to be registered separately. Once registered and listed in Tapplets Registry (ofc with checksum and so on) it is not planned to changed any data and upgrade is done within new PR to the Tapplet Registry. Only codeowner of a specific version can do it.
2. Tapplet's manifest, `tapplet.manifest.json`, is placed in the tapplet package (I'll add sample tapplet package structure to visualize it)
3. Before adding a tapplet to the Registry, CI will check if proposed package has correct structure, valid checksum etc. If does - next CI will generate `tappletsRegistry.manifest.json` file. So the contect hash can't be updated without checking an app before. Next version can be added with a new PR to the Tapplets Registry repo only by the codeowner.

CjS77 (approved):
LGTM

src/RFC-1101_Tapplet.md
```diff
@@ -0,0 +1,161 @@
...
+## Related Requests for Comment
+
+[RFC-1100](https://github.com/tari-project/rfcs/pull/134)
+
+## Description
+
+### Motivation
+
+The **Tapplet** is dedicated type of applet for the Tari blockchain and the fundamental part of the Tari Universe described in [RFC-1100](https://github.com/tari-project/rfcs/pull/134). The idea behind registered tapplets is to solve the class of vulnerabilities of the Dapp model known as supply chain attacks. This covers a wide range of attacks like: injecting malware code, XSS, stealing domains, social engineering attacks, etc.
+The solution proposed by Tari uses the tapplets Registry (another RFC will describe it) to register hashes of zip bundles of applications and sign them using the private key associated with Yat. The actual zip bundle can be hosted by https, but thanks to the registry commitment, the client can verify that the bundle hasn’t been tampered with.
```
stringhandler, line 62: Can you explain why a Yat is needed. IMO yats should not be involved

karczuRF: Thanks for that question. Basically Yat integration was just an idea, because as I know the Tari Aurora app already has Yat support, so it could be used as part of the Tari. But it's ok to skip it and not  involve Yat at all.

PR #138, opened 2024-03-22
feat: rfc-1102 tapplet registry

karczuRF:
The Tapplets Registry is the fundamental part of the Tari Universe. This RFC propose the registry structure as well as manifest file as list of registered and verified tapplets.


stringhandler (commented):
Possible con for NPM: There is a risk of dependencies introducing vulnerabilities (supply chain attack)

src/RFC-1102_TappletsRegistry.md
```diff
@@ -0,0 +1,263 @@
...
+
+- Github repository
+- Tari Blockchain - contract to store data onchain
+- Npm package registry (based on MetaMask Snaps)
+
+In next sections different scenarios are discussed as usage examples. Pros and cons of each solution can be found and summarized.
+
+## Github Repository
+
+Github repository is the solution which assumes that every tapplet's metadata is stored in a separate folder, each version in a subfolder. Tapplets themself are distributed az zip bundles, so the code is not stored in the folder. Tapplet Registry manifest file, named `tapplets-registry.manifest.json`, keeps metadata about verified and listed tapplets.
```
CjS77, line 72:
```suggestion
Github repository is the solution which assumes that every tapplet's metadata is stored in a separate folder, each version in a subfolder. Tapplets themself are distributed as zip bundles, so the code is not stored in the folder. Tapplet Registry manifest file, named `tapplets-registry.manifest.json`, keeps metadata about verified and listed tapplets.
```

src/RFC-1102_TappletsRegistry.md
```diff
@@ -0,0 +1,263 @@
...
+|     │  ├─ logo.svg
+|     │  ├─ background.svg
+|     ├─ README.md
+|     ├─ tapplet.manifest.json
+|  |- v1.0.1
+|  |- v1.0.0
+tapplets-registry.manifest.json
+```
+
+## Tari Blockchain
```
CjS77, line 89: nit: Usually "Tari Blockchain" refers to the Layer 1, since the L2 is not a blockchain `per sê`. Tari Network, or L2, is more unambiguous.

src/RFC-1102_TappletsRegistry.md
CjS77, line 72: If I understand this right:
All curated tapplets' source code will be stored together in a single monolithic Github repo. 
I guess this means that to add your app to the registry, you open a PR adding your code to the monolith. Is this right?

Edit: My question is answered below. I'll leave this comment in as an indication that it took a while to grok what this paragraph was saying.

src/RFC-1102_TappletsRegistry.md
```diff
@@ -0,0 +1,263 @@
...
+### Updating a tapplet version
+
+The upgrade process is almost the same as creating a new one, but with PR only changes need to be checked.
+
+### Deprecating/removing a tapplet
+
+It may (and pretty sure will) happen that a specific version is not recommended for use because of a bug or a business decision. In that case following options are considered:
+
+1. Mark the version as “deprecated” - special optional tag like “status” can be used
+2. Update the Tapplet Registry and remove the tapplet from the list available ones
```
CjS77, line 152: Ok, I asked about this in RFC-1101. 
So, there's a high degree of trust being placed in the maintainers of the registry repo. 

It would be good to make this trust model explicit somewhere.

CjS77 (commented):
A few nits. But LGTM

CjS77 (approved):

PR #139, opened 2024-03-25
feat: post exact copy of tokenomics blog post

stringhandler:
Reposts the tokenomics blogpost as an RFC. 

This is an EXACT copy. I am not the author of the original post. These are not my opinions.


CjS77:
Rather move this to tari.com's blog and we'll prep a more technical tokenomics doc with additional details around genesis block UTXOs for the RFC.

PR #140, opened 2024-04-02
feat: rfc-1104 confirming transactions

karczuRF:
**RFC-1104 Tari Universe: rules of confirming transactions.** 
The aim of this Request for Comment (RFC) is to describe the rules for confirming transactions by the user.


src/RFC-1104_ConfirmingTx.md
```diff
@@ -0,0 +1,158 @@
...
+
+   Transaction with presented data can be confirmed by clicking a button on a tapplet window, which appears or on top of a tapplet or is placed aside.
+
+5. Desirable from the point of view of readability and UX.
+
+   Simulate the transaction and summarize the difference of user holdings in case the transaction is executed. It can be done with transaction dry-run before submitting it to a validator node. For more details see [RFC-0350/TariVM](https://rfc.tari.com/RFC-0350_TariVM)
+
+#### Suggested solution
+
+As it was discussed, the security comes first, so definitely all transactions need to be confirmed before sending to the Tari Network. At the same time users shouldn’t be overwhelmed by the amount of information, especially not given just the “list of hashes” to sign, which are definitely not understandable for typical users. Therefore the idea is to show the transaction summary and quite a nice example of that is presented by [Radix Wallet](https://www.radixdlt.com/wallet).
```
stringhandler, line 98: "confirmed" is a term already used in blockchain, so for clarity, I suggest changing this to say "confirmed by the user"

stringhandler (commented):
In addition to this, I would like to add an extra level. 
Background: 
Due to approvals in Eth contracts, there are a number of attacks where the user is tricked into signing a transaction granting access to the malicious user to act on their behalf. For example an attacker could create a phishing site and then trick the user to sign approval for an ERC20 to the attacker's public key.

In Tari approvals and transactions happen differently, but I think there is still some things we can do to prevent the user from falling victim to these attacks.

One thing we could do is add a warning to the confirmation dialog if there is something suscipious about the transaction they are signing. We want this to be a clear warning that cannot be missed by the user, so we need to be careful of showing the warning too often. We could also disable the approval button for a number of seconds (like a countdown) to make sure the user can taken the time to see the warning.

This would also require the UI to have some understanding of what a dangerous transaction is. This gets a bit complicated, but I think that the warning can be mentioned in this PR and the implementation can be left to a later date.

stringhandler:
Please rename this to "User confirmation when submitting transactions"

PR #141, opened 2024-04-10
feat: rfc-1105 integrating wallet daemon into Tari Universe

MCozhusheck:
This RFC describes how Tari Universe will integrate wallet daemon

Storing private keys and confidential data is crucial part of Tari Universe. To ensure security we suggest to adapt [wallet daemon from tari-dan](https://github.com/tari-project/tari-dan/tree/development/applications/tari_dan_wallet_daemon) with some changes regarding communication where tapplet and wallet daemon communicate through IPC provided by Tauri framework.


stringhandler (commented):
Seems reasonable, but why not just use the JSON-rpc exposed by the wallet daemon already?

stringhandler (commented):
> Not sure how to describe this problem accurately, so I'll write it out and maybe we can substitute the correct terms later

We also need to make sure that the security mechanism and access granted to the tapplet cannot be distributed to a website and reused. 

For example, say the tapplet communicates with the wallet and is given a JWT granting it spending access. A malicious tapplet could send this JWT to a server and make requests from there. This must be prevented.

MCozhusheck:
JSON-RPC creates attack vectors where hostile actor could send requests to this rpc. By making direct communication with wallet daemon through IPC we keep JWT on the backend and just allow main view (through provider) to make calls and just return results to the tapplet. In this case tapplets needs to delegate calls to the wallet daemon to the provider which will ask users for permissions for listed actions that he will obtain from manifest.

PR #164, opened 2026-05-03
docs: tip-1: Tari Improvement Proposals

Kelketek:
## Description

This creates TIP-1, the first Tari Improvement Proposal, which establishes a process for creating Tari Improvement Proposals.

## Motivation and Context

Contained within the PR's contents.

## Dependencies

This PR depends on [this PR](https://github.com/tari-project/rfcs/pull/163) for its structure.

## How Has This Been Tested?

Manually.

PR #168, opened 2026-05-11
docs: charter v7 community feedback + structural changes

metalaureate:
## Charter v7: Community feedback + structural changes

Incorporates feedback from the May 2026 community comment period and reviewer feedback. Each change addresses specific concerns raised by community members and reviewers.

---

### Article II — Stipend baseline

**Problem:** A fixed XTM stipend with no stated value becomes either worthless or absurd as token price moves. No baseline to measure future changes against.

**Solution:** Initial stipend set at 100,000 XTM/month with a USD-equivalent reference cap of $4,000/month, calculated at a $0.04 forward-looking reference rate. The Council must publish its calculation method, review it quarterly, and any change requires the full amendment process with community ratification.

---

### Article IV — Council as active role + removal appeals

**Problem:** Nothing prevented a Council member from collecting a stipend while doing nothing. And if a member was removed, the Council was judge, jury, and executioner with no recourse.

**Solution:** Council service is now explicitly an active working role. Attendance, votes, recusals, and monthly activity summaries must be published. Removed members may propose a 3-person ad hoc Review Panel (no current Council members, no conflicted parties) to review the decision. The Council must adopt the recommendation or publicly explain why not.

---

### New Article VI — Core Contributor Program

**Problem:** "Sustained, verifiable contributions" was the eligibility standard for Council nomination, but there was no formal status, no binary in-or-out determination, and every individual's eligibility was arguable.

**Solution:** A formal Core Contributor Program with binary status — you either hold it or you don't. The Council admits contributors via a public Contribution Rubric with written explanations for each decision. Status can be revoked (with appeal rights). The inaugural Council must publish the rubric and admit the first Core Contributors within 90 days.

---

### Article VII — Anyone can be nominated, Core Contributors elect, Council members recuse

**Problem:** The Council confirmed or declined its own successors. Restricting nominations to Core Contributors only would create a closed loop. And Council members who also hold CC status could vote in their own elections.

**Solution:** Anyone with sustained, verifiable contributions can be nominated. Core Contributors are the electors. Council members who hold CC status must recuse from votes on their own seat. The Council defines the election procedure (public nominations, open discussion, recorded vote), but CCs make the selection decision.

---

### Article IX — Community ratification by Core Contributors

**Problem:** Ratification votes required a supermajority of "contributors who meet the Contribution Rubric threshold" — an unbounded, arguable standard.

**Solution:** Ratification requires a two-thirds supermajority of Core Contributors. Binary status, no ambiguity. The Council administers the program but cannot override the electorate.

### DIFF
--- a/src/TIP-0002_tari_community_charter.md
+++ b/src/TIP-0002_tari_community_charter.md
@@ -92,7 +92,7 @@ Tari Labs, the protocol's instigator, participates in the Council as a member, n
 
 The Council shall consist of five to seven members, drawn from the community and from enterprise, serving terms of twelve months. Initial terms shall be staggered so that no more than half of the Council's seats expire in the same period. No term shall be presumed to renew. At the conclusion of each term, the community may nominate the member to continue, or the seat shall pass to another.
 
-Council members shall receive a monthly stipend for their service. The stipend shall be denominated in XTM but subject to a publicly stated USD-equivalent reference cap. The Council shall publish the calculation method used to determine the stipend and shall review it no less than quarterly. Any material increase to the stipend, or change to the calculation method, shall require the charter amendment process and community ratification.
+Council members shall receive a monthly stipend for their service. The stipend shall be denominated in XTM but subject to a publicly stated USD-equivalent reference cap. The Council shall publish the calculation method used to determine the stipend and shall review it no less than quarterly. Any increase to the stipend, or change to the calculation method, shall require the charter amendment process and community ratification.
 
 ## Article III. Authority and Responsibility
 
@@ -116,7 +116,7 @@ A Council member shall demonstrate, through sustained conduct, the qualities the
 
 Council service is an active working role, not an honorary position. Members are expected to maintain regular availability, participate in scheduled Council deliberations, review proposals, vote when required, and contribute materially to at least one area of Council responsibility. The Council shall publish attendance, votes, recusals, and a brief monthly activity summary for each member.
 
-A Council member may be removed by majority vote of the remaining Council for sustained misconduct, prolonged unexplained absence, or conduct incompatible with this Charter. The removal and the reasoning behind it shall be made public. A removed Council member may request review by an ad hoc Review Panel composed of past Council members, past core contributors, or other community-recognized contributors who are not current Council members. The Review Panel may issue a public recommendation. The Council must either adopt the recommendation or publish its reasons for declining it.
+A Council member may be removed by majority vote of the remaining Council for sustained misconduct, prolonged unexplained absence, or conduct incompatible with this Charter. The removal and the reasoning behind it shall be made public. A removed Council member may request review by proposing three community members to serve as an ad hoc Review Panel. Proposed panelists must not be current Council members or have a direct stake in the outcome. The Review Panel may issue a public recommendation. The Council must either adopt the recommendation or publish its reasons for declining it.
 
 ## Article V. Decisions
 
@@ -134,7 +134,7 @@ The Council shall maintain a public Contribution Rubric defining the categories,
 
 Nominations shall be made publicly, with evidence of the candidate's contributions and a clear account of why they would strengthen the Council. The community shall discuss each nomination openly. The Council shall confirm or decline based on the consensus it observes. Every nomination, confirmation, rejection, and removal decision shall include a public written explanation referencing the rubric and the evidence considered.
 
-A rejected nominee may request review by an ad hoc Review Panel as described in Article IV. The Review Panel may issue a public recommendation. The Council must either adopt the recommendation or publish its reasons for declining it.
+A rejected nominee may request review by proposing an ad hoc Review Panel as described in Article IV. The Review Panel may issue a public recommendation. The Council must either adopt the recommendation or publish its reasons for declining it.
 
 The Council shall reflect the breadth of the protocol's needs: development, infrastructure, community, and enterprise. No single interest shall dominate.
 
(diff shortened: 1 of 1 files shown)


src/TIP-0002_tari_community_charter.md
```diff
@@ -92,7 +92,7 @@ Tari Labs, the protocol's instigator, participates in the Council as a member, n
 
 The Council shall consist of five to seven members, drawn from the community and from enterprise, serving terms of twelve months. Initial terms shall be staggered so that no more than half of the Council's seats expire in the same period. No term shall be presumed to renew. At the conclusion of each term, the community may nominate the member to continue, or the seat shall pass to another.
 
-Council members shall receive a monthly stipend of 100,000 XTM for their service.
+Council members shall receive a monthly stipend for their service. The stipend shall be denominated in XTM but subject to a publicly stated USD-equivalent reference cap. The Council shall publish the calculation method used to determine the stipend and shall review it no less than quarterly. Any increase to the stipend, or change to the calculation method, shall require the charter amendment process and community ratification.
```
Kelketek: There should probably be the initial values here, either directly in the charter or in some accompanying public declaration, just because we should start out of the gate knowing what the value is rather than having the initial council set their payout arbitrarily. Changes can then be compared against that base reference.

src/TIP-0002_tari_community_charter.md
```diff
@@ -126,13 +128,17 @@ Strategic matters, including significant spending, protocol changes, and amendme
...
+
+The Council shall maintain a public Contribution Rubric defining the categories, evidence, and example thresholds used to evaluate sustained, verifiable contributions. The rubric may include, without limitation, merged code, infrastructure operation, documentation, testing, security review, governance facilitation, community support, ecosystem development, and completed funded work.
 
-Nominations shall be made publicly, with evidence of the candidate's contributions and a clear account of why they would strengthen the Council. The community shall discuss each nomination openly. The Council shall confirm or decline based on the consensus it observes.
+Nominations shall be made publicly, with evidence of the candidate's contributions and a clear account of why they would strengthen the Council. The community shall discuss each nomination openly. The Council shall confirm or decline based on the consensus it observes. Every nomination, confirmation, rejection, and removal decision shall include a public written explanation referencing the rubric and the evidence considered.
```
Kelketek: Am I reading this correctly to mean that the current Tari Council determines who may succeed them? These requirements seem like something the council might use to evaluate Core Contributor candidates, and if CCs were the primary council electors, that would make some sense to keep here.

But writing in a specific mechanism by which council members can exclude successors, even one that seems like it should be an obvious requirement (no misconduct, contribution requirements) make these questions of judgement that the council can't be held accountable for, since they're the only ones with the power to make them.

src/TIP-0002_tari_community_charter.md
```diff
@@ -146,4 +152,10 @@ This obligation shall be the measure against which every future Council is judge
 
 This charter may be amended through the same process required for strategic decisions: open discussion, a Final Comment Period, and Council approval.
 
+Any amendment affecting Council powers, member selection, removal, stipend calculation, or amendment procedures shall require community ratification by a two-thirds supermajority of contributors who meet the Contribution Rubric threshold defined in Article VI, following the public discussion and Final Comment Period. The Council shall maintain a public register of eligible contributors.
```
Kelketek: This is an unbounded calculation that could result in the right of any individual community member to vote being litigated. There needs to be a more firm method of determining whether someone has voting rights or not.

I'd suggest writing that the Council shall establish a Core Contributor Program, and you could keep the notes about making a rubric, and judging against that for admission. The CCs could then be the electors, as mentioned in my earlier comment.

Kelketek (changes requested):

src/TIP-0002_tari_community_charter.md
PotR3288: The distinction between the Council and Core Contributors is unclear to me. @Kelketek , would you clarify what would constitute a Core Contributor's expected roles, criteria for selection, and mechanism of induction? How would those differ from those of the Council? Would the Council select Core Contributors and Core Contributors select the Council?

Kelketek: @PotR3288 Much of the specifics are to be set up by the council as they construct the CC program, but in brief, a CC:

* May have commit access to one or more repositories
* May be granted access to specific project resources (websites, social media accounts, etc) as needed to perform duties
* May be in charge of product direction/project management for one or more Tari subprojects
* Has voting rights for council members and may become a council member
* Shows consistent dedication to the project and good/skilled behavior (per the rubric), as a condition of their induction

A council member:

* Votes on [TIP](https://github.com/tari-project/rfcs/pull/164)s
* Decides on treasury disbursements with the rest of the members
* Sets project goals at a high-level
* Administers the CC program

Kelketek: Realized I didn't answer all of the questions. Here's further comment:

1. The Council will be appointed by Tari Labs
2. The Council will set up the CC program
3. Initial CCs will be inducted. It is expected that Tari Council members will also be inducted as CCs.
4. When the council terms are up, elections are held for council positions. Only CCs are eligible to run and vote.

PotR3288: Thank you. My concern with point 4 of the second reply is that if council members can be selected only from core contributors there will be a  prescribed path of enthusiast -> core contributor -> council which then requires all council members to have been selected by a previous council to become core contributors according to a rubric not yet defined.  

As you mentioned in today's Tari space, efforts are needed from all realms of experience. Could this process lead to a gate of sorts, keeping out talent from areas not typically considered? With the core contributor program and the council tightly coupled, could it become a centralizing closed loop?

As selection criteria for core contributors would be later determined by the council and not codified in the charter, is there a risk that governance could die on the vine?

metalaureate: @PotR3288 Good catch. I've updated Article VII. The candidate pool and the electorate are now two different things.

Who can be nominated: Anyone with sustained, verifiable contributions to the protocol (code, infrastructure, docs, community, ecosystem, the works). You do not need Core Contributor status to be considered for a Council seat.

Who votes: Core Contributors.

Who administers the CC program: The Council, via a public rubric with written justifications for every admission decision.

The CC program exists to formalize electoral rights, not to prescribe a career ladder. There is no required path of enthusiast → CC → Council. Someone who has done significant ecosystem work but never applied for CC status can still be nominated, discussed, and elected. The rubric is a gate for voting rights, not for candidacy.

This also addresses your concern about governance dying on the vine. Even if the rubric starts narrow (and it shouldn't, but even if it does), the candidate pool remains as broad as the protocol's contributor base. The Council can't accidentally lock out the people it needs most. @Kelketek

Kelketek: I like this change. Thanks, @metalaureate and @PotR3288 !

PR #174, opened 2026-06-11
feat: adds tip004 pow changes

SWvheerden:
Description
---
adds a proposed changed to change how the difficulty is calculated


src/RFC/MinoTari/TIP-RFC-MT-0004_MinoTari_PoW_Difficulty_changes.md
```diff
@@ -0,0 +1,157 @@
...
+
+| TIP             | [P-TIP-PROC-MT-0004](#/Process/MinoTari/TIP-Proc-MT-0004_MinoTari_PoW_Difficulty_changes.md)                  |
+|-----------------|---------------------------------------------------------------------------|
+| Title           | MinoTari PoW difficulty changes                                               |
+| Last Modified   | 2026-05-17                                                                |
+| Authors         | SW van Heerden                                                            |
+| Status          | Proposed                                                                  |
+| Type            | Process                                                                   |
+| Created         | 2026-06-11                                                                |
+| References      | Based on the Open edX community's [OEP] process. Forum discussion thread. |
```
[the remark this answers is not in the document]
Kelketek: I might suggest linking one or more previous TIPs/RFCs here-- especially any that might need to be updated if this is accepted.

Kelketek (commented):
@SWvheerden Small comment on the header, but otherwise this looks good to me. Exponential backoff as a mitigation to making any lane dominant seems straightfoward to implement and likely to resolve immediate issues.

src/RFC/MinoTari/TIP-RFC-MT-0004_MinoTari_PoW_Difficulty_changes.md
```diff
@@ -0,0 +1,157 @@
+# P-TIP-PROC-MT-0004: MinoTari PoW difficulty changes
```
Kelketek: @SWvheerden I was taking another look at this, and realized that this TIP should not be under `PROC` but under `RFC`, so this would be `P-TIP-RFC-MT-0004` (or perhaps some other number?) per [this table](http://localhost:3000/TIP-0001_tari_improvement_proposals.html#tip-types).

Accordingly the directory this is under should probably change as well before we merge and are stuck with it.

stringhandler (commented):
I don't think this will solve the problem.

Some faults off the top of my mind:
1. If you are behind on C29 blocks, multiple C29 blocks are hard to mine in succession.
2. If you interleave 2 algorithms, there is no penalty.

If anything you should adjust the target time for the other 3 algorithms when a C29 block has not been mined. This strategy is complicated and also flawed.

You would have to find a sequence in the chain's history where this proposal would have actually made a difference. 

From my experience with P2pool, trying to artificially limit POW does not have the desire effect, and often has the reverse of what you are trying to achieve.

src/RFC/MinoTari/TIP-RFC-MT-0004_MinoTari_PoW_Difficulty_changes.md
```diff
@@ -0,0 +1,157 @@
+# A-TIP-RFC-MT-0004: MinoTari PoW difficulty changes
+
+| TIP             | [A-TIP-RFC-MT-0004](#/RFC/MinoTari/TIP-RFC-MT-0004_MinoTari_PoW_Difficulty_changes.md)                  |
+|-----------------|---------------------------------------------------------------------------|
+| Title           | MinoTari PoW difficulty changes                                           |
+| Last Modified   | 2026-05-17                                                                |
+| Authors         | SW van Heerden                                                            |
+| Status          | Accepted                                                                  |
+| Type            | Process                                                                   |
```
Kelketek: Type is `Architecture`, per [the type definitions](https://rfc.tari.com/TIP-0001_tari_improvement_proposals#tip-types)

src/RFC/MinoTari/TIP-RFC-MT-0004_MinoTari_PoW_Difficulty_changes.md
```diff
@@ -0,0 +1,157 @@
...
+This creates a situation where one algorithm can disproportionately
+reorg blocks mined by other algorithms. That outcome undermines the
+original design goal of including multiple PoW algorithms for
+decentralization and hardware diversity.
+
+
+## Proposed Change
+
+Tari currently calculates target difficulty using LWMA (Linearly
+Weighted Moving Average) over the last 90 blocks. The LWMA uses:
```
Kelketek, line 95: @SWvheerden Could you make sure to add the change to the LWMA as discussed in the forum thread and approved by the council? I think we landed on 45, but Kinkajou should remember if neither of us does.

src/RFC/MinoTari/TIP-RFC-MT-0004_MinoTari_PoW_Difficulty_changes.md
```diff
@@ -0,0 +1,157 @@
...
+    **8 minutes**.
+
+This mechanism:
+
+-   Does not alter the underlying hash rate.
+-   Does not directly manipulate accumulated difficulty.
+-   Makes selfish mining exponentially more expensive for any single
+    algorithm attempting to dominate.
+-   Encourages natural interleaving of algorithms.
+-
```
Kelketek:
```suggestion
```

src/RFC/MinoTari/TIP-RFC-MT-0004_MinoTari_PoW_Difficulty_changes.md
```diff
@@ -0,0 +1,157 @@
+# A-TIP-RFC-MT-0004: MinoTari PoW difficulty changes
+
+| TIP             | [A-TIP-RFC-MT-0004](#/RFC/MinoTari/TIP-RFC-MT-0004_MinoTari_PoW_Difficulty_changes.md)                  |
+|-----------------|---------------------------------------------------------------------------|
+| Title           | MinoTari PoW difficulty changes                                           |
+| Last Modified   | 2026-05-17                                                                |
```
Kelketek: Please bump this upon addressing notes

Kelketek (changes requested):

src/RFC/MinoTari/TIP-RFC-MT-0004_MinoTari_PoW_Difficulty_changes.md
SWvheerden: I was treating that as a separate move, but we can do it here as well.

SWvheerden:
@m4r1m0 
Does your simulations run with the penalty classes or straight-up algorithms?

m4r1m0:
@SWvheerden edit: oops, my sim does use the old penalty classes 
`js/config.js: PENALTY_CLASS_MAP = { 0: 'randomx', 2: 'randomx', 1: 'sha3x', 3: 'c29' }`

I didn't update it yet for your latest change to the TIP

m4r1m0:
updated: https://github.com/m4r1m0/tari-diff-sim/tree/sw-penalty-per-algo

Results look better than the initial proposal now. Kinda rushed this, still need to run some more tests on the sim so I'll keep at it.

SWvheerden:
@m4r1m0 
I pushed up the initial Rust code for a ref
https://github.com/tari-project/tari/pull/7960

PR #185, opened 2026-07-11
feat: Core Contributor Program

Kelketek:
## Description

This Pull Request adds a TIP for the Core Contributor Program.

## Motivation and Context

The Tari Council is required to establish a Core Contributor program within 90 days of inauguration. This begins the work and provides a framework for the project, including requirements for induction, scope, and voting rights.

## How Has This Been Tested?

Locally. Render seems fine.


Kelketek:
Thanks, @claytonbittle . This is excellent feedback. I've pushed a change that should address most of your notes:

> 1. Ecosystem is missing as a category while the charter makes reference to it.

I've included your wording for Ecosystem CCs.

> 2. Admission mechanics conflict with charter Article VI.
The charter says the Council "shall admit contributors who meet the rubric's threshold and shall publish each admission decision with a written explanation referencing the rubric and the evidence considered." By contrast, this TIP has admission determined by a peer vote: five affirmatives are required, and a single 'no' defeats the nomination outright. Two issues: (a) it's a blackball where one incumbent CC can gatekeep the electorate indefinitely, with no way to override or appeal; and (b) it shifts admission authority from the council-against-published-rubric to peer unanimity. Suggested reconciliation: keep CC voting as the primary mechanism, but a 'no' triggers council adjudication against the rubric with the written explanation the charter already requires, rather than auto-failing the candidate.

I've taken your recommended solution, but have also stipulated that it be a supermajority. The council should need to feel very strongly about overriding CC judgement. Ultimately, if a CC is consistently blocking new CCs from being added for poor reasons, they could be removed under conduct reasoning per the involuntary removals clause.

> 3. The bootstrap veto deserves a fix before it's ever used.

The fix is inferred by the previously mentioned fix.

> 6. Wallet signature powers shouldn't follow the standard rights-expansion path.

I've added a note that this specific power must require an additional supermajority vote from the council.

> 7. The TIP defines who has standing but doesn't yet define how decisions flow.

I do think this is scope creep, and might be better addressed by a follow-up summary TIP or article for publication on a wiki in the future.

> (2) The PR description says 30 days but the charter (Article VI) says 90. (3) The timelines mixes business days, working days, and calendar weeks. It would be best to make it one unit throughout.

Fixed the description. Made the handling of days business days almost everywhere. The exception was elections, which I felt were better served by calendar dates.

>  IRV is specified, but staggered terms will produce multi-seat election batches, which need STV or explicit per-seat races.

I think this is actually more of an issue that needs to be addressed in either the charter or via one-off council action to stagger the seats. Currently, all six council members have started their terms at the same time, and would end them at the same time, which doesn't satisfy the staggered terms the charter mentions. We may need to have some special one-time extension election for a randomly selected set of half of the council, but that's outside the scope of this document.

Two sets of elections a year could do it or every other year if we want to do two-year terms. These are just ideas, not anything I'm hard recommending yet, but we should address them after the most immediate critical items are handled so it doesn't become an issue.

PR #188, opened 2026-07-17
feat: convert types

SWvheerden:
Description
---

Motivation and Context
---

How Has This Been Tested?
---


src/RFC-0110_BaseNodes.md
```diff
@@ -1,4 +1,12 @@
-# RFC-0110/BaseNodes
+| TIP             | [I-TIP-RFC-MT-0110/BaseNodes](#I-TIP-RFC-MT-0110/BaseNodes)               |
```
Kelketek: This document is missing the h1. Could you please add it?

src/RFC-0111_BaseNodeArchitecture.md
```diff
@@ -1,8 +1,17 @@
-# RFC-0111/BaseNodesArchitecture
+# I-TIP-RFC-MT-0111/BaseNodesArchitecture
```
Kelketek:
```suggestion
# I-TIP-RFC-MT-0111: BaseNodesArchitecture
```

Style of current TIPs is to use a colon and space rather than a slash.

src/RFC-0110_BaseNodes.md
Kelketek:
```suggestion
| TIP             | [I-TIP-RFC-MT-0110][#I-TIP-RFC-MT-0110-BaseNodes]               |
```
The second header should have just the identifier rather than a short title. The title's got a segment in the header here. Revised anchor is what the generated ID should be with colon and space-d title.

src/RFC-0111_BaseNodeArchitecture.md
```diff
@@ -1,8 +1,17 @@
...
+
+| TIP             | [I-TIP-RFC-MT-0111/BaseNodesArchitecture](#I-TIP-RFC-MT-0111/BaseNodesArchitecture)|
```
Kelketek: Same note here about the top column and anchor (and for the rest of the documents)

src/SUMMARY.md
```diff
@@ -2,84 +2,85 @@
...
-  - [RFC-0190: Mempool](RFC-0190_Mempool.md)
-  - [Tari-specific extensions to Mimblewimble](BaseLayerExtensions.md)
-    - [RFC-0201: TariScript](RFC-0201_TariScript.md)
-    - [RFC-0202: TariScript Opcodes](RFC-0202_TariScriptOpcodes.md)
-    - [RFC-0204: TariScript Examples](RFC-0204_TariScriptExamples.md)
-    - [RFC-0203: Stealth Addresses](RFC-0203_StealthAddresses.md)
-    - [RFC-0205: Hardware Transactions](RFC-0205_HardwareTransactions.md)
-    - [RFC-0250: Covenants](RFC-0250_Covenants.md)
-- [RFC-0303: The Tari Digital Assets Network](RFC-0303_DanOverview.md)
+- [Proposed]
```
Kelketek: mdbook is not happy with this revised structure. There might be some further tricks required to get the layout you're going for here. Looks like you will probably need to [make actual documents for segment intros](https://rust-lang.github.io/mdBook/format/summary.html#structure).

Kelketek (changes requested):

src/SUMMARY.md
SWvheerden: should be now

src/accepted.md
```diff
@@ -0,0 +1,5 @@
+# Accepted
+
+Proposals in this section have been reviewed and accepted by the community, but have not necessarily been
+implemented yet. Once the corresponding functionality has been built and merged, they move to the
+[Implemented](implemented.md) section.
```
Kelketek: Unless we want to move all accepted process TIPs to Implemented eventually, we should make a note that TIPs that don't require an implementation may stay in this category.

src/implemented.md
```diff
@@ -0,0 +1,5 @@
+# Implemented
+
+Documents in this section describe functionality that has been implemented in the Tari codebase. They form the
+authoritative specification of how the Tari network currently works, covering the base layer, its extensions and
+the applications built on top of it.
```
Kelketek: And here, we should make a paired statement to the other one-- that if a TIP is one that doesn't require a specific implementation, it remains in accepted.

Provided, of course, we decide not to mark all process and similar TIPs as 'implemented.'

Kelketek (changes requested):

src/implemented.md
SWvheerden: I would say if it does not require any specific implementation, it goes from accepted straight to implemented. 

If some Tips are accepted but are already in effect, but others require some changes before they go into effect, its going to cause confusion.

src/accepted.md
Kelketek: @SWvheerden Could you update the wording for this section (and the Implemented section) to closer match the definitions in TIP-PROC-1? Specifically about how items with no specific implementation are handled.

Kelketek (approved):
@SWvheerden Just one small note. If you address that, it's 👍 from me.

src/implemented.md
```diff
@@ -1,5 +1,5 @@
 # Implemented
 
-Documents in this section describe functionality that has been implemented in the Tari codebase. They form the
+Documents in this section describe functionality that has been implemented in the Tari codebase and or the in effect. They form the
```
Kelketek: I think this got mangled. Maybe "or proposals that are otherwise in effect"?

src/implemented.md
```diff
@@ -1,5 +1,5 @@
...
 authoritative specification of how the Tari network currently works, covering the base layer, its extensions and
```
Kelketek: "how the Tari network and project governance" might be good to specify here.

Kelketek (commented):
Just a couple of notes on the wording.

